PR #12, opened 2017-09-08
Refactor: map module, with service and component emitters

Lane:
Turned out it was pretty clear what should go in the service and what should stay in the component, so I went ahead and refactored the map module with a service.

I'm using the component for any input and output, and the service for any calls to modify the map.  Let me know if anything seems off.


src/app/map/map.module.ts
```diff
@@ -0,0 +1,20 @@
...
+import { MapboxComponent } from './mapbox/mapbox.component';
+
+
+@NgModule({
+  exports: [
+    MapboxComponent
+  ],
+  imports: [
+    CommonModule,
+    MapBoxModule.forRoot(
```
pjsier: This is a small thing, but do we actually need an API key, given we're using our own tiles (other than OpenMapTiles for now)?

Lane: I think we do, I tried to initialize without an API key and it crashed.  Though, I'm not sure if it was the `angular-mapbox` package I'm using, or if it was `mapboxgl` itself that wouldn't initialize without it.

pjsier: Weird, I just looked at the Mapbox docs and I guess it does, good to know! https://www.mapbox.com/help/how-access-tokens-work/

Lane:
I think it's good practice to keep the individual components separate from application logic, so that we have as few dependencies as possible in the individual components to encourage component reuse.  

Here's an article that gives an example (smart vs. dumb components): http://blog.angular-university.io/angular-2-smart-components-vs-presentation-components-whats-the-difference-when-to-use-each-and-why/

pjsier:
Perfect, that's what I was looking for! I'll take a closer look at that and merge in the meantime

PR #18, opened 2017-09-18
Fix Mapbox rendering on prod

### LINKED ISSUE #7: Production build breaks the map component
Something goes wrong with mapboxgl when running `ng build` with the `--prod` flag, and the map does not render.

pjsier:
Fixes #7 by making the changes described in that comment. The package `angular-mapbox` was also contributing to the problem, and we weren't really using it/it doesn't seem to be maintained so that's removed here. Adds typings for `mapboxgl` and `geojson`, and replaces the package functionality with the corrected types. Works with `ng build --prod` and then running the static files on localhost.

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -20,9 +20,12 @@
       "prefix": "app",
       "styles": [
         "styles.css",
-        "../node_modules/bootstrap/dist/css/bootstrap.min.css"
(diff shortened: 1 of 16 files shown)


Lane:
Great! I didn't know there were GeoJSON types we could pull in for typescript, so that's good to see.

PR #21, opened 2017-09-20
Basic implementation of data years

pjsier:
Added methods and changed layers to support a `dataYear` property on the main app component, as well as a `censusYear` property for the decennial census. We'll probably need to change parts of this going forward, but it sets up a framework for the year selector component.

The most changes are in the `style.json` file, and it's pointing at a different bucket where I deployed the tiles from the pending PR in `eviction-lab-etl`. If this basic setup works, I think we can merge that as well (although Travis isn't cutting it for that deployment, so we'll probably have to do it semi-manually for now)

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -33,7 +33,7 @@
             <span class="legend-end">{{legend[1][0]}}</span>
         </div>
     </div>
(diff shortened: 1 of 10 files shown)


Lane:
Looks great!  Definitely a good base to provide functionality for the year slider.

PR #22, opened 2017-09-20
generic select component for map ui

Lane:
generic select component to reduce code duplication and simplify styling


src/app/app.component.html
```diff
@@ -15,15 +15,19 @@
         (zoomChange)="setMapZoomLevel($event)">
     </app-zoom-control>
     <div class="map-ui">
-        <app-layer-select 
-            [selectedLayer]="activeDataLevel"
-            [layers]="dataLevels" 
+        <app-ui-select
+            class="layer-select"
+            [labelProperty]="'name'"
```
pjsier: Can this (and the one below it) be simplified to `labelProperty="name"`? That worked for me locally

Lane: Yep! Good catch, just updated.

PR #24, opened 2017-09-21
Disable touch rotate

pjsier:
When I was playing around with the mobile prototype, I accidentally rotated the map a bit while zooming. I don't know if we really have a use case for rotating, so it made sense to disable it for touch gestures. There's also [an option to disable rotation for right-click and drag](https://www.mapbox.com/mapbox-gl-js/example/disable-rotation/) which could be worth considering too


Lane:
Yep, makes sense to disable it to me.

PR #27, opened 2017-09-25
Deploy prototype branches from travis

pjsier:
Adding a shortcut for deploying prototypes to S3 as long as the branch starts with `prototypes/`. It didn't work initially, but not sure if Travis allows for changing deployment settings from a non-master branch

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,10 @@ deploy:
   script: scripts/invalidate-cache.sh
   on:
     branch: master
+- provider: script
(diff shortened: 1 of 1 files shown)


Lane:
Good idea! Should we just merge this straight to master?

pjsier:
@Lane that makes sense to me! I can delete this PR and make a new one branched off master if that works?

Lane:
Sure, you could probably just edit this one so the base is `master` instead of `development`.

PR #31, opened 2017-10-03
Map data updates

pjsier:
This removes the census year functionality from the app, cleans up styling, and attempts to update bubble properties in addition to choropleth. I think I might be running into a bug on that last one though (which is why I updated Mapbox GL JS).

I've tried doing both `setPaintProperty`, and now even `setStyle` to change the styles for the entire layer, but for some reason the bubble styles won't update. I have a feeling it's a bug in Mapbox GL where zoom + property functions aren't changed on style updates even if property functions are, but I'm stuck on it either way. The code is in `setLayerDataProperty` in the map service.


pjsier:
Forgot this earlier, but even though the year slider is buggy, the bigger update on this is for moving away from the decennial layers and to one layer with all years. I'm fine with removing some of the extra features from it, but because of that I branched `prototypes/map-seed-data` off of this

PR #32, opened 2017-10-03
Add vertical option to UI slider, replace zoom

pjsier:
Added a vertical option to the UI slider that does a CSS transform on the elements (as well as some additional styling changes). Also converted the `currentValue` property to an input with getters and setters to handle the two-way relationship with the values on zoom.

I wasn't sure the best way of breaking out the styles between the component and the parent so I did most of the style updates for vertical in the component, but that gets a little wonky with the transform.


pjsier:
Because of an error I was running into, I made `getStepValue` public. It only became an issue because zoom has a two-way relationship with the data unlike year, but I'm guessing we'll want to clean that up

src/app/map-ui/ui-slider/ui-slider.component.scss
```diff
@@ -72,4 +72,20 @@
     transition: transform 0.2s ease;
     transform: scale(1);
     
+}
+
+.el-slider.vertical {
+    transform: rotate(270deg);
```
Lane: Instead of rotating the element to create the vertical version, I think it might be better to have a fixed width and a fluid height, so that it fills the height of it's parent container.  That way styling the component is easier, as we could specify the height styles instead of specifying width styles that will actually reflect the height of the component.

pjsier: Yeah, the transform approach was easier at first, but it'd be more complicated to maintain, works for me!

PR #53, opened 2017-10-04
WIP: Cursor tooltip after hover delay

pjsier:
Still figuring out the specifics of this interaction, but wanted to open the PR to show where it's at currently. This is based off of the conversations on having a more reduced tooltip instead of the larger one on the bottom of the sceen


Lane:
Good start! I think for the interaction we need to do one of the following to avoid the tooltip from blocking mouse events to the map:

  - Tooltip always follows the mouse. (something like the tooltips here: https://www.nytimes.com/interactive/2015/05/03/upshot/the-best-and-worst-places-to-grow-up-how-your-area-compares.html)
  - Tooltip fades out and disappears when the user starts moving the mouse

pjsier:
Either one of those options sounds better to me, but I'm having a little trouble figuring it out in RxJS. I'm guessing there's a relatively simple way of doing it there, does anything come to mind?

PR #55, opened 2017-10-04
Change map container to ViewChild

### LINKED ISSUE #33: Mapbox component should use ViewChild element
Currently the mapbox component uses `id="map"` in the component template, and expects `container: 'map'` to be passed to the component in the map configuration.

We should not expect the `container` to be specified in the map config, and instead have the mapbox component grab the container element by:

  - Removing the `id` attribute and adding `#map` on the element
  - using `@ViewChild` to get the `nativeElement` in the component
  - Passing the `nativeElement` to the `MapService.createMap`, which should create the map in the provided `nativeElement`

These changes will:
  - Allow for multiple maps on the page (no element id conflicts)
  - Remove the need to specify `container: 'map'` in the map config.
  - Ensure the map is injected into the correct element / Fix map overlaying karma tests

pjsier:
Closes #33

### DIFF
--- a/src/app/map/mapbox/mapbox.component.html
+++ b/src/app/map/mapbox/mapbox.component.html
@@ -1,1 +1,1 @@
-<div id="map"></div>
+<div #map id="map"></div>
--- a/src/app/map/mapbox/mapbox.component.html
+++ b/src/app/map/mapbox/mapbox.component.html
@@ -1,1 +1,1 @@
-<div #map id="map"></div>
+<div #map></div>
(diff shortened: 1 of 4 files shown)


src/app/map/mapbox/mapbox.component.html
```diff
@@ -1 +1 @@
-<div id="map"></div>
\ No newline at end of file
+<div #map id="map"></div>
```
Lane: Should pull the ID attribute off of this element, just so there are no conflicts if we want multiple map components at some point

pjsier: Good call! updated

PR #62, opened 2017-10-06
Improved hover performance, reduced layers

pjsier:
I was looking around at some best practices and examples of using Mapbox GL and vector tiles, and I found a hover implementation that's significantly faster https://github.com/ccusa/Disaster_Vulnerability_Map/blob/master/js/mapbox.js#L206

Instead of adding an additional layer for each source, we can just have one hover layer that's blank GeoJSON and update the source with the GeoJSON-like feature returned from the event. Here's some screen recording gifs for comparison

Current State:
current-hover

With this PR:
improved-hover

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -148,18 +148,10 @@ export class AppComponent {
     if (this.hover_HACK > 0 || !this.platform.isMobile()) {
       this.hover_HACK = 0;
       this.hoveredFeature = feature;
-      const hoverLayer =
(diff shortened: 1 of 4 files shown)


pjsier:
The caveat for this is that sometimes the hovered geography will be split between tiles, so it'll only highlight the part within the tile. That's also something we could address in the tile creation step though by specifying `--no-clipping` in tippecanoe, but that has its own drawbacks

Lane:
Awesome! This is much better.

PR #63, opened 2017-10-06
Add UI progress bar component

### LINKED ISSUE #52: UI Progress Component
A component that is used to display progress of a task, or plays an animation for actions of  indeterminate time.

  - [x] Accepts input of progress amount (0 - 1)
  - [x] Accepts boolean input for `loading`, which determines if the component is actively loading.
  - [x] Displays bar that fills a % corresponding to `progress` value
  - [x] Displays indeterminate animation if no progress is given.
  - [x] Default style / theme variables

pjsier:
Closes #52. Relatively straightforward, but after seeing how often the map loads, I think using a bar rather than spinner is good because it shows up a lot and this is visible but not too intrusive.

Also, I used the `:host` selector to make it completely full-width, but I can use a container element that in if it makes more sense.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -24,10 +24,12 @@
             (ready)="onMapReady($event)"
             (zoom)="onMapZoom($event)"
             (featureClick)="onFeatureClick($event)"
-            (hoverChanged)="onFeatureHover($event)">
+            (hoverChanged)="onFeatureHover($event)"
+            (render)="onMapRender($event)">
         </app-mapbox>
     </div>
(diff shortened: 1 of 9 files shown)


Lane:
A few things we might want to adjust / discuss later: 

  - the conditions for when it shows up.  we might want to add a couple hundred millisecond delay so it doesn't pop up for extremely short operations.
  - using the progress version for under the header, even if we fake it.  I could see the indeterminate animation being distracting across the whole page.

The base component looks great though, merging.

Issue #42, opened 2017-10-03
Map Zoom Slider


Lane:
[Functional Spec](https://docs.google.com/document/d/1_20EvW4w73O-yjt22THLxGh1b6rTe-cyF2ZZXfLnUvk/edit#heading=h.13jv3h7l1jmc)

Todo:

  - [x] Implement using Mapbox default +/-

Lane:
Updated to switch implementation to mapbox controls.

pjsier:
Can we close this after #67?

PR #67, opened 2017-10-10
Remove zoom slider and replace with buttons, add geolocation control

pjsier:
Based on some of the UX feedback, it sounds like we might be removing the zoom slider and replacing it with the standard +/- buttons. A button for geolocation has also been discussed, so adding it in as well (it won't show up on staging sites because S3 sites are over HTTP)


pjsier:
Fixes #42

PR #68, opened 2017-10-10
WIP: Prototypes/low zoom layers

pjsier:
This is an interaction that's been talked about, but not confirmed on whether we want to use it. Opening a PR so we don't forget we have this branch out there.

`MapLayerGroup`s now have a `minzoom` property that sets the minimum zoom level for them to be displayed. If a layer with minzoom is being viewed, and the users zooms out past the minzoom, auto switching of layers is turned back on to handle the currently displayed layer.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -47,7 +47,7 @@
                 class="layer-select"
                 labelProperty="name"
                 [selectedValue]="activeDataLevel"
-                [values]="dataLevels" 
+                [values]="selectDataLevels"
                 (change)="setGroupVisibility($event)">
             </app-ui-select>
             <app-ui-select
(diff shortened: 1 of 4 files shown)


Lane:
I was checking this out yesterday, makes sense to me.  Should we do the same for max zoom as well?  If you're looking at a zoom level 10 for example, the state layer is not going to show you anything.

pjsier:
Let's put a hold on this. Social Explorer seems to be able to do it with their tileserver, so I'm sure we could find a way of tweaking the parameters to allow for block group view on a nation-level scale

pjsier:
@Lane I'm guessing we still want to do this, right? I can close this and re-implement it since it's quick

Lane:
Yeah, I think we still want to do this?  It makes sense if we can't give a good low-level view of the smaller geographies.

pjsier:
Closing this and re-opening another

PR #76, opened 2017-10-11
Switch map boolean to fix progress

pjsier:
This seems like it shouldn't make a difference, and we'll need to verify, but it looks like using the more general `loaded()` (which I don't know why I didn't use in the first place) resolves the issue. To test on a deployed site, I updated one of the prototypes with this branch http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/low-zoom-layers/


Lane:
Cool, looks good on the prototype link!

PR #80, opened 2017-10-12
Enable IE polyfills

pjsier:
Just making sure we support the world's best browser 💯


Lane:
For flexbox, using workarounds is preferred: https://github.com/philipwalton/flexbugs

If there are too many or if it's too time consuming and a polyfill fixes everything, then let's go with a polyfill.

PR #81, opened 2017-10-12
Feature/ui hints

### LINKED ISSUE #69: UI Hint Directive
A directive that can provide a hint over text content.

  - [x] Directive pulls value from one of the following attributes, in order of priority: `hint`, `aria-label`, `alt`, or `title`
  - [x] Displays the hint value in a tooltip on mouseover
  - [x] Displays the hint value in a tap for touch devices
  - [x] Has a style / visual indicator to show there is a hint for the contained content

Lane:
A basic wrapper for bootstrap tooltips to display a "help" icon with a clickable hint.

image

For word hints we'll just use the tooltip directive in `ngx-bootstrap`

Closes #69

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -39,7 +39,6 @@
             (change)="setMapZoomLevel($event)">
         </app-ui-slider>
         <div class="map-ui">
-            
             <app-ui-select
                 tabindex="4"
                 class="eviction-select"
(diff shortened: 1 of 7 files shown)


src/app/app.component.html
```diff
@@ -39,6 +39,7 @@
             (change)="setMapZoomLevel($event)">
         </app-ui-slider>
         <div class="map-ui">
+
```
pjsier: Small thing but we should probably get rid of this whitespace

PR #82, opened 2017-10-13
Fix auto switch layers by re-ordering data levels

pjsier:
Not sure why this hadn't been triggered earlier (potentially was refactored), but the initial on-change event for the UI-select was getting triggered after render, which then called `setGroupVisibility` and disabled auto-switching. We should probably have the layers in descending order of size anyway, so I made states the first layer as well as the default layer so that the initial load doesn't trigger a change event and it seems to be working


Lane:
Ah, makes sense.  Thanks for tracking that down!

PR #84, opened 2017-10-17
Refactor/modules

Lane:
Refactor into `app-map` and `app-data-panel` components to clean up `app.component` logic.


pjsier:
Not sure if you wanted to make this ready for review yet, but it looks great! One thing I'm wondering about that's semi-related is the amount of methods in `MapService` that are really just duplicating calls to `map` functions like `setLayerStyle`, `setLayerFilter`,  and `setZoomLevel`.

I think the abstraction makes sense for things like `setLayerDataProperty`, but for those earlier methods it just seems duplicative. Would it make sense call it`mapService` rather than `map` on the map component and call some `map` methods directly? `MapboxComponent` has a somewhat similar approach, but I don't think we would need to add `map` as a property directly

Lane:
Yeah, I think there is definitely some additional refactoring to be done.

My thinking for the mapService was that it would handle all calls / interactions to the MapboxGL API and would not expose the mapboxgl.Map for direct use, except to the mapbox component to setup event bindings.  My main reason for doing this was just in case we wanted to switch to a different map implementation later (*highly unlikely*), we would only need to rewrite the MapService.

Hold off on merge for now, I will clean this up a bit more today.

pjsier:
ah gotcha, if that's the case it works for me!

PR #90, opened 2017-10-19
Prototype for Mapzen search geocoding

pjsier:
Added Mapzen geocoding with the search service. Had to modify the predictive search component to use the `waitMs` parameter and debounce searches as well as emit changes to search text to parent component. Also fixed inconsistences in naming in style.json (block groups was `block-groups` everywhere but there), to search layer selection work. 

Not sure if the way the app component is changing the layer on the map (`setDataLevelFromLayer`) is the best, but it's a placeholder for now. Also not sure about the best way of testing the service, other than stubbing requests.


Lane:
Looks great!  I'd be fine with changing how we set the location on the map to center / zoom instead of bounding box.

PR #91, opened 2017-10-19
Support splitting tiles by census, multi year data queries

pjsier:
Using `forkJoin` to combine multiple HTTP requests and merging the objects returned. Work in progress because it's difficult to test without the data setup, and I think it's pretty clear at this point that we'll need to split at least on a decennial basis, so I'll add that to the ETL repo before marking this ready.

UPDATE: Updated to handle switching layer sources based on decennial census year. The `updateCensusSource` function is pretty specific because updating the style is expensive, and it didn't seem like something we would use outside of this use case. Search service now accepts optional boolean `multiYear` which determines whether all sources are queried and merged in response. Also updates Mapbox GL JS to potentially resolve bug mentioned in `setLayerDataProperty` comment.


Lane:
Looks good to me!  As this relates to the graph, we'll have to bring in the search service and make a call to `getTileData` with the `multiYear` flag.  How should we get the  lat/lng coordinates for the feature to call `getTileData`? 

  - grab the center point based on the `north`,`south`,`east`,`west` properties on the feature
  - is there some lat / lng value available on the feature that I'm missing?

pjsier:
@Lane yep, we'd just add the search service, and the return value from the geocoder has the coordinate array in `feature.geometry['coordinates']`, so we could just pull it from that the same way it gets added for the map

pjsier:
Adding context from Slack conversation here. Pulling from the geocoder works if a feature is added to the graph through the data panel search box, but not if it was added by clicking it on the map. If a feature was added that way, we can just call `getTileData` with `multiYear` as true, and the coordinates would just be the center of the bounding box

PR #92, opened 2017-10-23
Georgia Prototype

Lane:
Todo:

  - [x] Center map on Georgia
  - [x] Update tilesets to only include Georgia
  - [x] Add bounds to map so you can't move away from Georgia
  - [x] Turn off choropleths by default (poverty rate is first available choice)
  - [x] Two style variations


Lane:
I think this is ready to merge.  I removed all of the Georgia specific stuff, so it should be the same as the current development map / data.

PR #101, opened 2017-10-27
Add tooltip when label not visible

pjsier:
Checks both if a feature has been rendered generally with `queryRenderedFeatures` as well as whether it's visible by checking the layer `text-opacity` property, and only displays a tooltip if the label isn't visible either way


Lane:
Nice! Looks good.  One small fix:  could you make it so the tooltip hides once the mouse is moved off the map?  Currently it sticks:

image

pjsier:
Good catch! Should be fixed now
tooltip

Lane:
Looks good!

PR #103, opened 2017-10-27
Restrict layers by zoom level

pjsier:
Updated version of #68. `MapLayerGroup`s now have a `minzoom` property that sets the minimum zoom level for them to be displayed. If a layer with `minzoom` is being viewed, and the users zooms out past the `minzoom`, auto switching of layers is turned back on to handle the currently displayed layer.

### DIFF
--- a/src/app/data/data-levels.ts
+++ b/src/app/data/data-levels.ts
@@ -10,6 +10,7 @@ export const DataLevels: Array<MapLayerGroup> = [
             'states_bubbles',
             'states_text'
         ],
+        'minzoom': 0,
         'zoom': [0, 7]
     },
     {
@@ -21,6 +22,7 @@ export const DataLevels: Array<MapLayerGroup> = [
             'counties_bubbles',
             'counties_text'
         ],
+        'minzoom': 0,
         'zoom': [7, 9]
     },
     {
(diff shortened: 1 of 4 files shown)


Lane:
Great!  I have a few updates for this one:

  - Should we show all of the levels in the select, and just grey out the ones that are unavailable at the current zoom level?  I wouldn't want someone top open that menu zoomed out, and think that states / counties are the only available options.
  - I think the zoom levels might need adjusting?  I zoomed it and selected block groups, but the outlines were showing me census tracts.  Example below:

Image

pjsier:
Agreed about graying out, but I'd also want to check with Abby on that because I think she preferred not displaying them at all. And for the census tracts, I think that's just the full label for block groups, so the name is TRACT.1, Census Tract TRACT

Lane:
Ah, makes sense!  Should we keep a small border at the lower zoom levels?  If there is no choropleth highlighting, you can really tell the boundaries that the circles represent.

pjsier:
Hmm, I could go either way on keeping a border. On the one hand it could be confusing, but on the other hand the borders will be for the aggregated geometries, and now that we're restricting the zoom levels the layers won't be visible at all for a larger portion of the map. Maybe we could also consider some way of making it more apparent which layer you're on?

PR #109, opened 2017-10-30
Feature/graph tooltips

Lane:
Base tooltip implementation on graphs.  We might need to adjust how tooltips show on line graphs to prevent overlap of similar values.

image


pjsier:
This looks great overall, but for the line graphs should we only show a tooltip for the closest line rather than all lines at the X position? That's at least how it is in the latest mockup we got. Since it's just displaying the value, it wouldn't matter if two lines overlap exactly

screen shot 2017-10-30 at 12 47 28 pm

PR #112, opened 2017-10-31
Pull data across years for clicked features

pjsier:
Currently data outside of the displayed tile's layer is only pulled for features added through predictive search. This adds a call to `getTileData` with a point guaranteed to be inside a feature in response to feature click events


Lane:
Great! Thanks for doing this.

PR #116, opened 2017-10-31
File download form

pjsier:
For the initial implementation the only thing left is some kind of loading gif and tests, but wanted to get this PR in since the email form is similar. I think that can probably be a separate component, but let me know if this setup makes sense!


Lane:
Yep, this is looking good to me.  I think I will probably create a separate dialog component for the email modal as well.

PR #132, opened 2017-11-07
Add initial mailto link

pjsier:
I'm assuming this should still work regardless of the routing implementation, but let me know if it'll cause issues.


pjsier:
Closing and including in general social share PR

Issue #121, opened 2017-10-31
Email Share


Lane:
Link the "Email" icon button in the bottom row of the data panel to a `mailto:` link that pre-populates the message with some text and a link to the current view.

pjsier:
Pending task here is having some kind of popup notification for users that click `mailto` without having a default mail client set up.

pjsier:
Closed by #133 and #174

Issue #119, opened 2017-10-31
Twitter Share


Lane:
Add click binding on Twitter icon in the footer row to open a pre-populated "compose tweet" dialog that includes the link to the current view.

Dialog can be open with a link like:
https://twitter.com/intent/tweet?text=Hello%20world

pjsier:
Is this also closed by #133? Also, is it an issue that people won't be able to access any of the sharing items unless they've selected a feature on the map? If that's intentional it's fine with me, just realized it though

pjsier:
Closed by #133

Issue #118, opened 2017-10-31
Facebook Share


Lane:
Link the facebook icon in the footer row to a facebook share dialog.

- [x] Add open graph metadata to page (title, description, localization)
- [x] Use dynamic image if possible, or fallback to static image

[Open Graph Tags](https://developers.facebook.com/docs/sharing/best-practices#tags)
[Example FB code](http://www.fbrell.com/saved/2779dc018c325d85d650a3b723239650)

pjsier:
Should we consider this closed by #133 and make a separate issue for social media metadata?

pjsier:
Closed by #133

PR #133, opened 2017-11-07
Add basic mailto and social share functionality, metadata

pjsier:
Progress on #121, #119, and #118. Adds basic functionality for a `mailto` link as well as a directive for opening social share popups. Creates the `getCurrentUrl` method, which I'm assuming shouldn't be dependent on the router, but is needed because `window` can't be used in the template. 

Facebook's social share doesn't work locally because it actually tries to pull metadata from the supplied URL, but it works when provided an external link. I also added some sample metadata from the preview site that we can change. As for adding a dynamic image, it looks like it's pulling from the page directly, so that might be more trouble than it's worth (this StackOverflow answer seems to confirm that https://stackoverflow.com/a/29515024)


Lane:
Looks good! Definitely more trouble than it's worth for a dynamic image, thanks for looking into that.

PR #134, opened 2017-11-07
Feature/deeplinking

Lane:
Deep linking implementation, from the README:

The following options can be set through app routes or URL parameters:

  - Locations
  - Map Bounds
  - Eviction Type
  - Choropleth Highlight
  - Geography Layer

## Routes

Routes for the app are as follows:

http://{{BASE_URL}}/:locations/:year/:geography/:type/:choropleth/:bounds

With the following options

  - `:locations`: a list of locations separated by `+`.  Each location is formatted as `{{layerId}},{{x}},{{y}}`.  Set to `none` for no active locations.
    - e.g. `states,257,381+counties,230,385` would activate Illinois and Kearney County
  - `:year`: a year anywhere from 1990 to 2016
  - `:geography`: `states`, `counties`, `cities`, `zip-codes`, `tracts`, or `block-groups`. 
  - `:type`: an ID for the corresponding `BubbleAttribute` (`er` or `efr`)
  - `:choropleth`: an ID for the corresponding choropleth `DataAttribute` (`p` or `pr`, more to come)
  - `:bounds` a comma separated bounding box for the map to zoom to `{{west}},{{south}},{{east}},{{north}}`

Data must be set through routes in this order `:locations/:year/:geography/:type/:choropleth/:bounds`.  If you want to set only one of these parameters, use the URL Parameters.

## URL Parameters

Map settings can also be set through URL parameters instead of routes, with the following format:

http://{{BASE_URL}}/link;locations={{locations}};year={{year}};geography={{geography}};type={{type}};choropleth={{choropleth}};bounds={{bounds}}

All settings are optional, so you could set only the year, eviction rate bubbles, and bounds with:

http://{{BASE_URL}}/link;year=2015;type=er;bounds=-92.52,38.25,-86.53,41.76

### DIFF
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -9,6 +9,8 @@ import * as inside from '@turf/inside';
 import * as bbox from '@turf/bbox';
 import 'rxjs/add/observable/forkJoin';
 import * as _isEqual from 'lodash.isequal';
+import * as polylabel from 'polylabel';
+
 
 import { MapDataAttribute } from '../map-tool/map/map-data-attribute';
 import { MapLayerGroup } from '../map-tool/map/map-layer-group';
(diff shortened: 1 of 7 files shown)


src/app/map-tool/map/map/map.component.ts
```diff
@@ -77,7 +81,7 @@ export class MapComponent implements OnInit {
   /** Available layers to toggle between */
   @Input() layerOptions: MapLayerGroup[] = [];
   /** Toggle for auto switch between layerOptions based on min / max zooms */
-  @Input() autoSwitch = true;
+  @Input() autoSwitch = false;
```
pjsier: Was this leftover from testing, or do we need it for the reasons listed below?

src/app/data/data.service.ts
```diff
@@ -22,25 +23,97 @@ export class DataService {
...
+    }, '');
+  }
+
+  /**
+   * Gets an array of values that represent the current route
+   */
+  getRouteArray() {
+    const locations = this.activeFeatures.map((f, i, arr) => {
+      const xy = this.getFeatureXY(f);
+      return f.properties['layerId'] + ',' + xy.x + ',' + xy.y;
```
pjsier: Would it be easier to just have the center point latitude and longitude of a feature here? My worry is that X and Y are zoom-dependent, so if there are any other uses to this than just pulling the data it could be safer to use the lat lon coordinates

src/app/data/data.service.ts
```diff
@@ -22,25 +23,97 @@ export class DataService {
...
+    }).join('+');
+    return [
+      (locations === '' ? 'none' : locations),
+      this.activeYear,
+      this.activeDataLevel.id,
+      this.stripYearFromAttr(this.activeBubbleHighlight.id),
+      this.stripYearFromAttr(this.activeDataHighlight.id),
+      this.mapView ? this.mapView.join() : null
```
pjsier: Not sure if it'll cause the map to jump around, but could we simplify the coordinates here to 2 or 3 decimal places? The URL looks a lot more unwieldy as is

src/app/data/data.service.ts
Lane: Yeah, that's what I was thinking initially.  You had suggested using X/Y values because it's shorter and less processing, which made sense to me.  Are there concerns that we wouldn't get the proper feature using X/Y?

src/app/data/data.service.ts
Lane: Yeah, makes sense to me.  Will update.

src/app/data/data.service.ts
pjsier: Ahh whoops, my bad! If this was in reference to the exports, that library you sent me before cleared up any issues with using lat lon. And if it's just for getting tile data then no, but not sure if this would play into anything else

Issue #152, opened 2017-11-10
Fix feature geometry from getTileData


pjsier:
Right now `getTileData` is only looking at one tile, which is good for performance, but means that for features like states that are larger, the returned geography is often much smaller than the actual shape. We need an equivalent to `getUnionFeature` in the map service for getting tile data

pjsier:
So I'm going around in circles a bit on this one and wanted to get some options I've thought about down. To clarify, I thought this was only a problem with features that hadn't been added from a click or geocoding, but it looks like since `getTileData` isn't modifying the geometry it's a problem with any feature added there.

We can solve the problem for features added from search by adding the `bbox` property to the feature, but for features that we're just getting from `getTileData` for the first time (like through route params) here's what I've thought about so far.

- Query a lower zoom level for tiles to get the entire feature
  - Not sure how to handle places like California where the state is large enough to be cut off for almost all tile sizes
- Reverse geocode with Mapzen to get feature
  - Doesn't seem foolproof (did a test run with zip codes and it didn't return anything), and would rather not potentially add to our bill there
- Store bounding box for each feature stored in locations in the route params instead of just a center point
  - Goes back to original problem with getting features in tiles where the center of a bounding box might not necessarily be a point within the feature we're looking for
- For the cases where this would be an issue (route params), either have a static, per-layer zoom level whenever something like `zoomToFeature` is called, or switch to that layer and pull the bounds as soon as it loads
  - Seems like the only option I can think of at the moment, but also feels a bit hacky

Lane:
Just processing this a bit, but want to make sure I understand the issue:

- This is only an issue when trying to get a feature / data from route params
- The problem is that when it queries for tile data, it might not include the full set of geometry (like if a state has islands?)

pjsier:
So I thought it was only an issue when we're getting data from route params, but really it's any time we use `getTileData`. The code that's related is here https://github.com/EvictionLab/eviction-maps/blob/development/src/app/data/data.service.ts#L242

The problem is that when it queries for tile data, it only gets the part of the feature in the tile at zoom 10 (so for California, we're pulling a small square in the middle of the state). In `mergeFeatureProperties`, we merge the properties, but not the geometry. We also can't just merge the geometry, because it's still only pulling from whatever is in the small zoom 10 tile. 

Let me know if any of that doesn't make sense! Still trying to make sure I'm looking at it right

pjsier:
[...]

PR #153, opened 2017-11-10
Click card header to view on map

### LINKED ISSUE #149: Click on location name in card to zoom to it

pjsier:
Closes #149. I noticed the issue in #152 while making this, but I think it can be addressed separately.

card-header


pjsier:
If that boolean works then I think that's better than having the input! I'm only using it for that class, but I didn't think of using the `observers` property. I'll update that

PR #154, opened 2017-11-10
Add notice for hidden layers in select

### LINKED ISSUE #143: Layer select dropdown updates
Update layer select to only display available layers based on zoom, but if any are hidden it should have some indication that more layers can be found on zoom

image

pjsier:
Closes #143. Instead of just hiding the layers that can't be accessed or graying them out, whenever layers are hidden the message "Zoom for more" displays at the bottom of the select. Tried to modify as little as possible since UI select is used everywhere. 

Some other things to note:
* I took out the `max-height` for UI select because some of the layers aren't visible without it. If this was for mobile though I'm happy to change it back
* I added `cursor: pointer` to select items to make it more explicit that they can be clicked

layer-select

### DIFF
--- a/src/app/ui/ui-select/ui-select.component.html
+++ b/src/app/ui/ui-select/ui-select.component.html
@@ -17,5 +17,6 @@
         {{getLabel(value)}}
       </a>
     </li>
+    <li *ngIf="bottomLabel" class="bottom-label"><a class="dropdown-item">Zoom for more</a></li>
   </ul>
 </div>
--- a/src/app/ui/ui-select/ui-select.component.html
+++ b/src/app/ui/ui-select/ui-select.component.html
@@ -17,6 +17,6 @@
         {{getLabel(value)}}
       </a>
     </li>
-    <li *ngIf="bottomLabel" class="bottom-label"><a class="dropdown-item">Zoom for more</a></li>
+    <li *ngIf="bottomLabel" class="bottom-label"><a class="dropdown-item">{{bottomLabel}}</a></li>
   </ul>
 </div>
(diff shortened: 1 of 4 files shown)


src/app/ui/ui-select/ui-select.component.html
```diff
@@ -17,5 +17,6 @@
         {{getLabel(value)}}
       </a>
     </li>
+    <li *ngIf="bottomLabel" class="bottom-label"><a class="dropdown-item">Zoom for more</a></li>
```
Lane: Should `{{bottomLabel}}` be in the dropdown-item, instead of `Zoom for more`?  Just in case we want different messages in other dropdowns.

src/app/ui/ui-select/ui-select.component.scss
```diff
@@ -16,7 +16,8 @@
       box-sizing: border-box;
       min-width: 140px;
       padding:0;
-      max-height: 190px;
```
Lane: This is mostly for mobile, to make sure the content of the select dropdowns don't go off of the screen.  Happens mostly for year selector.  Could you remove the max height in a media query instead?

src/app/ui/ui-select/ui-select.component.html
pjsier: Ah whoops, that's a typo, I'll fix that

src/app/ui/ui-select/ui-select.component.scss
pjsier: Makes sense! I'll update that

PR #159, opened 2017-11-13
Move location cards into map

### LINKED ISSUE #115: Cards overlapping select inputs
Move `app-location-cards` from `app.component` into `map.component` and adjust z-index accordingly so that cards do not sit above select inputs.

pjsier:
Fixes #115. I'm just having the events on location cards bubble up to the `map-tool` component, and injecting the `DataService` there. Let me know if there's a better way of handling any of that, and I'm requesting a review, but I'm pretty sure some of these PRs are going to conflict with each other

### DIFF
--- a/src/app/map-tool/map-tool.component.html
+++ b/src/app/map-tool/map-tool.component.html
@@ -79,15 +79,10 @@
       (selectedBubbleChange)="updateRoute()"
       (selectedChoroplethChange)="updateRoute()"
       (selectedLayerChange)="updateRoute()"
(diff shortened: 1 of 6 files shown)


Lane:
Looks good, a couple things to fix up still:

  - The cards need to be outside of the `map-ui` div, so that they are fixed to the side of the screen (see screenshot)
  - I'd like to keep `MapToolComponent` as the only component with a dependency for `DataService`.  `MapComponent` has an Input for year already that you could use for the location cards, and you'd just need to add an `activeFeatures` input.

image

PR #171, opened 2017-11-15
Position progress bar over ui-select

pjsier:
With the UI select components on the map getting moved up, they're currently in front of the progress indicator which makes it less visible. Fixed here


Lane:
We'll be moving this to the top (above the header) as it is in the mockups soon, so maybe we should just do that now instead of adjusting z-index?

image

pjsier:
Good call! Closing this

PR #173, opened 2017-11-15
Set one fixed progress bar listening to events

### LINKED ISSUE #172: Update progress indicator to match mockups
Progress indicator should be at top of app screen overall. I was also thinking of adding an issue to use the progress indicator after someone's selected a search feature but before the data for it is loaded. Should we have it have a fixed position at the top level component and then listen for loading events from child components?

Mockup:
image

pjsier:
Closes #172. Use one main progress bar for both map and data service loading events. Right now, I don't like setting `dataService.isLoading` manually each time, but given the chain of observables and different places they can be called I'm not sure what the best way of managing that would be.

progress

### DIFF
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -31,6 +31,7 @@ export class DataService {
   autoSwitchLayers = true;
   mapView;
   mapConfig;
+  isLoading = false;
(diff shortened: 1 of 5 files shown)


Lane:
Looks good! I think setting `DataService.isLoading` manually works for now.

Issue #187, opened 2017-11-20
Map Flickering


Lane:
Map styles flicker on chrome when moving, zooming, hover.  Not 100% sure of the cause yet, but I think it might have something to do with updating map styles.

pjsier:
Looks like this was fixed in https://github.com/mapbox/mapbox-gl-js/releases/tag/v0.42.1, they actually mention in the fix that it's worst in maps with many sources

PR #190, opened 2017-11-24
update angular, update ngx-bootstrap

Lane:
Figured it'd be good to update sooner than later.  A few notes:

The `scripts.bundle.js` did not generate the first time I ran `ng serve`, so mapbox didn't load.  It has worked on every subsequent `ng serve`, so maybe try again if it doesn't work. [Issue for reference](https://github.com/angular/angular-cli/issues/8371)

Updates the local version of `@angular/cli` to `1.5.4`, so be sure to [update that](https://github.com/angular/angular-cli#updating-angular-cli).

Side note:  #187 is no longer happening for me after this update?  Not sure if the update fixed it, different chrome version, or something else.


pjsier:
Thanks for the notes! On the point about Chrome I've noticed that too, and I'm thinking it might be a Chrome update?

PR #195, opened 2017-11-27
BrowserStack setup, e2e scaffolding

pjsier:
Set up BrowserStack with local testing and added some additional e2e tests. There are some issues in BrowserStack with earlier versions of Safari, so ideally we'd test before 10, but that's what's working for now.


pjsier:
I'm hesitant to include the `e2e`tests as a mandatory step for the build to finish, but this at least gets the scaffolding for more browser tests, BrowserStack Local, and lets them run on each Travis build so that we can check out the recordings

Lane:
Awesome! I'm fine with the e2e tests not being mandatory for now.

PR #196, opened 2017-11-27
Tweak select transition, judgment spelling

pjsier:
Just noticed some minor quirks with the transition for the last map UI dropdown that I wanted to fix, as well as a "Judgment" label to update

select-transition


Lane:
Ah man, I've got to stop doing that! Looks good. 👍

Issue #70, opened 2017-10-10
Add translation json


Lane:
Work with Abby's content matrix to build `en.json` and `es.json` language files.

I think using [ngx-translate](https://github.com/ngx-translate) over [Angular's i18n](https://angular.io/guide/i18n) might be preferred.  `ngx-translate` will allow us to look up translations for dynamic values (like our UI select values), where Angular i18n would require us to create static templates for each component.

PR #197, opened 2017-11-28
i18n base implementation

Lane:
Base setup for language switching.  Todo:

- [x] setup so language JSON is loaded w/ ngx-translate
- [x] add handler to the language select component so that selected language is loaded from the corresponding JSON file

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -337,6 +337,16 @@
         }
       }
     },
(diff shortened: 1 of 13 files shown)


Lane:
This should be good to merge in.  Provides the functionality to complete #70

PR #219, opened 2017-12-01
Keep state borders visible

### LINKED ISSUE #144: Increase state outlines on all zooms
Not sure if this should be managed with the OpenMapTiles layers or the census layers, but state borders should be more visible for all layers

pjsier:
Closes #144. UX/design feedback on this would be good before merging, but I moved the state borders to the top, but still sourced from OpenMapTiles so we don't have to worry about code changes. I also removed the `maxzoom` property and increased the thickness bit on zoom

state-outlines

### DIFF
--- a/src/assets/style.json
+++ b/src/assets/style.json
@@ -3946,11 +3946,7 @@
           ]
         },
         "line-blur": 0.4,
-        "line-dasharray": [
-          2,
-          2
-        ],
-        "line-opacity": 1
+        "line-opacity": 0.8
       }
     }
   ]
--- a/src/assets/style.json
+++ b/src/assets/style.json
@@ -1146,48 +1146,6 @@
       }
     },
     {
-      "id": "boundary_state",
-      "type": "line",
-      "metadata": {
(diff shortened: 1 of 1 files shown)


Lane:
This looks good, I think Noele had mentioned wanting to turn off boundaries for other countries earlier in the project.  Could we do that and still use the OpenMapTiles layer, or should we use our own state boundaries layers?

pjsier:
Ah I forgot about that. I'll make an update to this now, it shouldn't be difficult to just switch the source

pjsier:
Just updated so that it uses the tile data source for states. I removed other boundaries in general, took out an extra US layer that wasn't getting used, and changed the `style.json` name because I noticed it was still "Natural Earth" and why not

Lane:
Looks good to me! Should I merge now or do you want to wait for design feedback?

pjsier:
I think it's fine to wait. I pushed it to the prototype link here http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/state-borders

Issue #247, opened 2017-12-04
Keep detail layers visible all the way to `maxZoom`


Lane:
Place details (roads, etc) should be visible on map at hight zooms to avoid white screen at high zoom levels.

pjsier:
@Lane was this for block groups disappearing or roads disappearing?

Lane:
Ahhh sorry, I was misreading the issue.   Will update!

Issue #246, opened 2017-12-04
Switch map font to "Metropolis Medium"


Lane:
Switch from Akkurat to Metropolis

Issue #236, opened 2017-12-04
Turn on county labels at higher zoom


pjsier:
Currently the county labels appear too early, so we should increase the minimum zoom where they display

PR #250, opened 2017-12-04
Multiple map style updates

### LINKED ISSUE #227: Increase road visibility at higher zooms
Right now it's not possible to look at cross streets to orient yourself while looking at geographies, which will definitely come into play with block groups. Given the map layer we're using is really minimal, it might be worth raising the transparency significantly

pjsier:
Closes #227,  #236, #246, #247. Changes map fonts to Metropolis, increases visibility of roads at higher zooms, keep roads on map at high zooms, limit county layer labels.

Also updated the stroke width and opacity to be 0.8 (except for states which have 1 for both) because not being able to see border lines at low zooms for cities in particular was mentioned as an issue. And removed `minzoom` restriction on bubbles for the city layer because it was confusing when it's the only layer with this restriction

Set this as a prototype because it'll be easier to review all the changes on a demo link:
http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/map-style-updates


pjsier:
For the block groups issue, I think it's mostly because streets overlap the block group boundaries a lot of the time, and the boundaries were already faint. 

I think one way to address the issue with cities would be to add some kind of default gray background for shapes, but it would really only help for cities because with all other layers they cover basically the entire map. 

The darkened highlight on hover makes sense. Should we just have it be the same color as the current stroke, but retain the width so that it's thicker?

Lane:
**Block groups:**
Maybe we could increase border width for block groups, and/or drop transparency on roads so they are a more transparent white, allowing boundaries to show through.

**Cities:**
I think a background could help.  I think adding markers for cities would make sense too.  Maybe small dots for small cities, and larger markers for major cities?

**Hover:**
Same color / thicker border make sense to me.

Also wondering if there's any value in boosting stroke width for all boundaries when no choropleth is selected.

pjsier:
I like the idea for block groups and boosting stroke width overall, but if the markers for cities are dots I think that's going to get confusing given we already have a set definition for what a circle is on the map. Could we see if the thicker border for cities works first?

Lane:
Yeah, for sure.  Good point, circles would be a bad choice.  I still think markers could be helpful at the low zooms though, just with a better choice of marker.  But yeah, let's see what it looks like with the thicker border and go from there.

pjsier:
Updated the hover stroke color and stroke widths in general, let me know what you think!

PR #251, opened 2017-12-05
Add ui-toast for location max select

### LINKED ISSUE #248: Handle maximum locations added
Show a dialog message when the user tries to add a fourth location.  Copy TBD, for now use: "You have selected the maximum allowed locations.  Dismiss one of the locations and try again."

pjsier:
Closes #248. Adds a toast notification when a user tries to select a location when they already have 3 location cards. Also adds one if they try to select the same location twice. Currently have it in the top right corner so we can reuse the general toast notification, but if it's important to have it closer to the cards we can probably add a function for dynamically setting the position (rather than just having that as an input.

location-cards

### DIFF
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -173,7 +173,7 @@ export class DataService {
    * Adds a location to the cards and data panel
    * @param feature the feature for the corresponding location to add
    */
-  addLocation(feature) {
+  addLocation(feature): boolean {
     if (this.activeFeatures.length < 3) {
       const i = this.activeFeatures.findIndex((f) => {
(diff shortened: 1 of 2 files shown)


Lane:
Looks good! I think it might make sense to display these in the lower left corner of the map, but should wait on some design feedback before moving.

One small update, Abby posted some text in #uxdesign to use for the message:

> Maximum limit reached. Please remove a location to add another.

Can we switch to that instead of the placeholder message.

pjsier:
Makes sense! Updated that, and I'll update the location once we hear back

Lane:
Merging in for now, will adjust position later if we are advised to do so.

PR #253, opened 2017-12-05
Add census demographics to dropdown

### LINKED ISSUE #202: Include other census variables in dropdown
The data for multiple census variables is ready, so we can go ahead and add it to the dropdown

pjsier:
Closes #202. I added everything that hadn't been included yet (only percentage for renter occupied households because I remember that being how it was described), and updated some of the translations. Research mentioned they wanted "Hispanic/Latinx" for that category, changed "Percent" to "%" in one, and added "Race" or "Races" to the last two for clarity.

Also, should we remove the "name" property from `MapDataAttribute` now that it's pulled from `langKey`?

### DIFF
--- a/src/app/data/data-attributes.ts
+++ b/src/app/data/data-attributes.ts
@@ -81,6 +81,198 @@ export const DataAttributes: Array<MapDataAttribute> = [
         [20, 'rgba(37, 51, 132, 0.9)']
       ]
     }
+  },
+  {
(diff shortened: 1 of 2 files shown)


Lane:
Great!

We could probably drop the name attribute when we do some refactoring.  I think it's fine to leave it in for now though, just so there is a default value before the translate service becomes available.

Issue #150, opened 2017-11-10
Location cards outline state


pjsier:
Look into highlighting the borders of any selected feature in location cards. 

**Note:** not sure how to handle this for location cards selected from different layers. The hover highlight is a separate layer, so we could potentially just leave the outlines there even if the current layer doesn't match. Also, look into making the highlight colors match the card colors

Lane:
Wonder if we could also add a marker?  Then the outline disappears when they switch layers, but the marker could remain.

pjsier:
It could work, but I wonder if that would get confusing with the bubble layers on top of it? Also, realizing that this will probably be dependent on #152

pjsier:
Thinking through this now, this could be more trouble than it's worth because of the issues in #152. Even if we add a bounding box to the URL, setting the outline would require getting the actual shape which we can't store in the route params

pjsier:
The only way I can think to do this easily is to create an S3 bucket that has one GeoJSON file for each feature, organized by layer and GEOID (because FIPS alone will be duplicated with zip codes and tracts). Then we'd pull the GeoJSON for the shape if it's not available. It might be worth considering how much this feature is worth it, but if it's a priority I think we can make it work

Lane:
Could we add additional "highlight" layers to the map, then filter based on name / parent location in the active features?

pjsier:
The problem is if the object isn't in the visible layer/features we would have no way of getting its shape. If we wanted to only show selected objects that are in the current map layer that could work though

pjsier:
I think it could make sense to just have another function that waits for onZoomEnd and adds any features that are currently visible? And maybe we just don’t worry about features in different layers that haven’t been toggled yet

PR #295, opened 2017-12-15
Highlight on selected card features 

pjsier:
The basics of this feature are working, implementing #150, but there are a couple quirks in behavior that we would need to decide how to handle before merging this.

The big question is how we should handle displaying shapes for features whose layers haven't loaded yet. Should we only display the outlines of features in the currently selected layer? The reason for this is if the layer isn't visible, on page load it will pull the small square for the zoom 10 tile (shown below). We could alternatively just try to ignore anything that's only loaded from `getTileData`, and wait until it uses `queryRenderedFeatures` to get the shape

Here's an example of how it's currently working (you can see some of the quirks, like the small square being loaded for Iowa):

location-outlines

### DIFF
--- a/src/app/map-tool/map/map.service.ts
+++ b/src/app/map-tool/map/map.service.ts
@@ -4,6 +4,7 @@ import { BehaviorSubject } from 'rxjs/BehaviorSubject';
 import 'rxjs/add/operator/distinct';
 import * as bbox from '@turf/bbox';
 import * as union from '@turf/union';
+import * as polylabel from 'polylabel';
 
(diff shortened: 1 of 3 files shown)


pjsier:
Just wanted to flag that the blue highlight doesn't really show up when a choropleth layer is applied

pjsier:
Made some updates so that the boundaries default to bounding boxes if the actual boundaries are unavailable, and then progressively update in detail on zoom. Also added a white outline to the boundaries to handle the overlap with blue choropleths in particular. Here's an example of how this will work, but overall I think this is ready for review now

highlight-lines

Issue #290, opened 2017-12-15
Rotate cards out instead of throwing error on selection past 3


pjsier:
It was mentioned at the meeting that people would have preferred for the first card to be dropped (with some kind of animation) and a new one added if a user clicks on a map feature when they already have 3 cards selected.

This makes sense, but it's also worth considering how/if this would affect the "add a location" data cards and other parts of the tool

Lane:
General consensus is that this is not a good thing.  Leaving prototype available here: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/card-swap

PR #309, opened 2017-12-19
Add card on click, then update data

### LINKED ISSUE #299: Card not displayed when clicking
Not sure if this was a network issue or what, but one of the cards would not show in the Princeton meeting when Matt was clicking on it.

I think we should adjust how the card is added so that it shows up immediately, and then the cards fill out as the Observables emit the data.

pjsier:
Closes #299. Changes `addLocation` into an add or update, adds clicked features to card locations immediately, then loads additional data later. It's working great overall, the only issue is that it is triggering a transition on the location cards on update as well as add.

I can't figure out where to prevent this, but I think if we get the sorted list of GEOIDs and check for equality (was initially thinking length being the same, but that wouldn't work with #290) we can verify if an event was an update rather than add. Where should I fix that?

### DIFF
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -170,21 +170,24 @@ export class DataService {
   }
 
   /**
-   * Adds a location to the cards and data panel
+   * Adds or updates a location to the cards and data panel
    * @param feature the feature for the corresponding location to add
    */
   addLocation(feature): boolean {
+    // Process feature if bbox and layerId not included based on current data level
+    if (!(feature.properties.bbox && feature.properties.bbox)) {
+      feature = this.processMapFeature(feature);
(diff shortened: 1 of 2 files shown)


pjsier:
Fixed the error mentioned above by just updating `properties` and `geometry` rather than the whole feature so a full state change isn't triggered for the animation

Lane:
This is working great!  However, now with the cards showing right away there is a noticeable delay for the outline of the area to show up.  Is that because it is waiting for the full updated geometry?  Could we immediately add the highlight based on the geometry of the clicked location?

pjsier:
It looks like the performance issue only showed up on Firefox, and even then wasn't all the time. Merging since there's not much we can do about that

PR #329, opened 2017-12-20
Update README, add code of conduct

### LINKED ISSUE #323: Open source notes
Update `README.md` to be more general, include brief description and instructions for installation and contributing. Move most of the current README content to `docs/`. Add a code of conduct.

Sort of related, but I'm also gonna add a DEPRECATED note to the descriptions of the old data repos like eviction-lab-tileset

pjsier:
Closes #323. Figured it would be better to do this now while we have some time. I moved the technical docs to `docs/components.md`, but given we aren't really planning on exposing this as a standalone package it could make sense to just get rid of it rather than try and keep it up to date?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,149 +1,19 @@
+# Eviction Lab Map Tool
+
 [![Build Status](https://travis-ci.org/EvictionLab/eviction-maps.svg?branch=master)](https://travis-ci.org/EvictionLab/eviction-maps)
 
-# EvictionMaps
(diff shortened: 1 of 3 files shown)


Lane:
Looks good, the technical docs are good in `docs/components.md` for now.  I had intended on finishing that up so it was there for reference, just in case someone needed to get up to speed on how the components work... but I forgot haha.  We can leave it for now, I'll either update it or remove it before launch.

PR #335, opened 2017-12-21
Refactor for rankings

Lane:
Still need to refactor toast and loading into a service so they can be accessed from multiple components.


Lane:
Should be ready for merge

PR #338, opened 2017-12-22
Switch debounce order to fix loading

### LINKED ISSUE #336: Loading indicator doesn't show as expected
Noticed this while refactoring, the loading indicator doesn't show as expected.

If I zoom or move around where some geography the loading indicator doesn't show.  Sometimes it shows momentarily, but it seems inconsistent.

no-loading

pjsier:
Fixes #336. Switching the order of debounce and distinct seemed to be enough. I retried your example and it works
loading

### DIFF
--- a/src/app/map-tool/map/map/map.component.ts
+++ b/src/app/map-tool/map/map/map.component.ts
@@ -269,7 +269,7 @@ export class MapComponent implements OnInit, OnChanges {
     this.updateMapData();
     this.map.isLoading$
       .distinctUntilChanged()
-      .debounceTime(200)
+      .debounceTime(500)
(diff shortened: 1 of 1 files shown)


Lane:
Hmm.. I am getting a constant flickering of the loading indicator with this.

load-flicker2

pjsier:
That's weird... I'm not getting that on Firefox or Chrome, so not sure what's going on there

Lane:
Yeah, it is weird.  Doesn't happen on Chrome for me either.  Only happens on Firefox, and only when I load the page an don't touch anything.  If I start interacting with the map it seems to go away.

Issue #297, opened 2017-12-16
Rework file download dialog


Lane:
There was a request to add additional information to the file download dialog to clarify what is included in the files and also potentially adding more formats (CSV, DOCX).  I'm thinking a good approach for this is to have a 3-step dialog with something like:

> Note: someone else should write the copy, this is just to show structure.

Step 1: Select type of export
---
**Spreadsheet**
Export a spreadsheet of `{{locations}}` for the years 2000 - 2016 including eviction, census, and ACS data.

**Document**
Export a document that includes and overview of evictions in `{{locations}}` and issues related to evictions. 

**Presentation**
Export a presentation with slides for `{{locations}}` including maps and graphs.

> If user selects an option with multiple file types (e.g. CSV, and XLSX) go to step 2.  If there is only one file type, go to step 3.

Step 2: Select file type
---
**Excel (.xlsx)**
Download the spreadsheet for Microsoft Excel.

**Comma Separated Values (.csv)**
Download the spreadsheet in comma-separated format.

Step 3: Download your file
---
Your export is being created...

> File begins downloaded when ready and dialog is dismissed.

I'm also leaning towards not using a form here, the user just clicks on the type of file and is moved to the next step.  Needs discussion with the team before implementing.

pjsier:
I like this a lot overall, but it seems like we don't need to make it into multiple steps if the only difference in step 2 is Excel vs CSV. I think we could just have those as two separate options (with similar descriptions), and keep the overall structure as-is

Lane:
Good point, could definitely get rid of step 2.

pjsier:
I'm remembering some conversations about including text at the top indicating what data the export would be for? I think it was something along the lines of "Generate a file export for Colorado, Michigan, and California data between 2000 and 2016." Making a note here so we don't forget

PR #339, opened 2017-12-22
Update download modal, add docx

### LINKED ISSUE #293: Re-add Word document export?
Need to confirm

pjsier:
Closes #293, #297. Still could use some tweaking, but makes it closer to what we're looking for

screen shot 2017-12-22 at 11 39 58 am


Lane:
Good for now.  I was thinking we could break them up by "document", "spreadsheet", and "presentation" (as mentioned in #297), then adding buttons for the various formats under those categories instead of having a form with a check box for each format.  That will need some design / UX review though, so I'll merge this in now as a starting point.

PR #344, opened 2017-12-26
WIP: Change ROH to renter homes

### LINKED ISSUE #283: Use "renter homes" instead of "renter occupied households"?
We're using "renter homes" instead of "renter occupied households", and I think it sounds a lot better and is also shorter which helps. Do we want to change this across the board?

pjsier:
Closes #283. The implementation itself is done other than translation, but marked WIP because we want to get feedback on the change first


Lane:
Renter homes is approved, is this good to merge?

pjsier:
Yep! Good to go

Issue #211, opened 2017-11-30
Add language to route as query parameter


Lane:
Language should be the first parameter on routes so that shared links are in the appropriate language.

pjsier:
Based on conversation in #346, using query params for this

PR #346, opened 2017-12-26
Add lang to map tool route params

pjsier:
Progress on #211. Only including for the map tool for now since I'm not sure the best way of incorporating it in there

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -91,15 +91,15 @@ export class AppComponent implements OnInit {
       year: 2016
     };
     const defaultViews = {
-      map: '/none/2016/auto/none/none/-136.80,20.68,-57.60,52.06',
-      rankings: '/evictions/United%20States/0/evictionRate',
-      evictors: '/evictors'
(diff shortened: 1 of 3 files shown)


Lane:
I was thinking we could use `queryParams` on the route for this instead of having it as a mandatory part of the URL.  (see: https://angular-2-training-book.rangle.io/handout/routing/query_params.html)

Might make sense to switch selected locations and choropleth to `queryParams` too, seeing as how they are optional?

Not ready to merge this one yet.  Will either need to add `:lang` to the `evictions` and `evictors` routes, or remove `:lang` from the map tool route and listen for the `lang` queryParam changes in `app.component` and switch the language when it is set or changes.

pjsier:
My preference would definitely be query params, so if that works for you we can close this and I'll make a comment in the issue about that. I also like the idea of moving locations and choropleths to query params too, but I'm guessing that would be a bigger change and would probably make more sense in a different issue?

Lane:
Sounds good.  Agree a separate issue for locations / choropleth is better left for another issue.

Issue #249, opened 2017-12-04
Adjust graph tooltip implementation


Lane:
To help with accessibility, place the value of the line / bar alongside the corresponding option in the legend instead of in a tooltip.

The graph should show an indicator for the area that was tapped / hovered.

pjsier:
The feedback for this is included in #425, so closing

PR #348, opened 2017-12-26
Add legend display of graph value

pjsier:
Progress on #249. Not 100% sure what is in mind here, but I took a stab at updating the legend with the current values. It disappears if nothing is hovered over for the line graph, and shows the currently selected year's value for the bar graph.

The styling isn't the greatest here, but I figured I'd at least make an initial attempt

screen shot 2017-12-26 at 1 33 03 pm
screen shot 2017-12-26 at 1 33 15 pm


Lane:
Cool, pretty much exactly what I had in mind.  Will merge in and mark #249 as feedback needed.

PR #353, opened 2017-12-28
Support for dashed null line

pjsier:
Not removing data for null years, using dashed lines instead. Dependent on https://github.com/EvictionLab/angular-d3-graph/pull/8


pjsier:
This works once the tweak in https://github.com/EvictionLab/angular-d3-graph/pull/10 is merged

Issue #356, opened 2017-12-28
Move locations and selected choropleths to query parameters


pjsier:
Based on conversation in #346

PR #361, opened 2017-12-29
Lang, locations, choropleth, and type as query params

### LINKED ISSUE #211: Add language to route as query parameter

pjsier:
EDIT: Closes #211, #356.

screen shot 2017-12-29 at 9 52 58 am


pjsier:
It looks like the language select issue was because we don't have any input for setting the current language, only change events, so I fixed that. The remaining issue is that "Auto" layers seem to lag on switching languages that were set on page load, but I'm guessing that's a separate issue. If it is, then this should be ready to go

screen shot 2018-01-02 at 2 45 30 pm

PR #367, opened 2018-01-04
Generalize geocoder functions, add Mapbox

### LINKED ISSUE #363: Switch to Mapbox geocoding API
Mapzen is shutting down, so we'll need to move this to Mapbox geocoding

pjsier:
Closes #363. Generalizes the search providers so that any specific geocoder needs to implement the `SearchSource` interface. Also adds Mapbox geocoder. This will make it easier to switch out providers. I added [OSM Names](http://osmnames.org/) as another example, and it could be a potential option since it might be cheaper given that it only does cities, counties, and states. It's through OpenMapTiles

**NOTE:** The Mapbox geocoder doesn't seem to have counties, but since that's not an incredible long list (3,007), we could potentially modify the search source to also go through a static list of counties to produce results.

### DIFF
--- a/src/app/ui/location-search/location-search.component.spec.ts
+++ b/src/app/ui/location-search/location-search.component.spec.ts
@@ -11,7 +11,6 @@ describe('LocationSearchComponent', () => {
   let fixture: ComponentFixture<LocationSearchComponent>;
   const searchServiceStub = {
     queryGeocoder: () => {},
-    getLayerName: () => {},
     query: ''
   };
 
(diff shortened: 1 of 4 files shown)


Lane:
Looks good! We will definitely want counties in there.  Should I merge this and add a separate issue for that, or do you want to add it to this PR?

pjsier:
@Lane that'll be a bit more work, so I think we're good to merge this for now. I'll make a separate issue for adding counties into search

PR #369, opened 2018-01-05
Disable body scroll on select open

### LINKED ISSUE #188: UI Select: Turn off page scrolling when menu is open

pjsier:
Fixes #188. Adds `SelectService` which tracks when UI select components are opened. The `blur` event fires after a new select is opened (if you have one open and then click on another), so the `SelectItem` interface tracts the label as well as the state to only mark the state `false` when the same element is closed. There's probably a better way of tracking it with an `ElementRef` rather than the `label`, but it works for now.

Also had to attach `overflow-y` to the body element rather than `app-root` because the `overflow-y: hidden` property on `app-root` isn't actually working. That might be a separate bug for the other styles that are trying to prevent scroll before items are selected.


Lane:
Ah I'm sorry about this one, I added the issue so I wouldn't forget about it but didn't outline the desired functionality very well.

We're looking for functionality that mimics the built in browser `<select>`.  As seen here (https://codepen.io/Lane/full/qppKRQ/)

- User can scroll the select contents all the way to the bottom, if they keep scrolling inside of the menu when it hit's the bottom, the page doesn't scroll.
- If the user scrolls outside of the menu, the page scrolls and the menu closes.

Should be able to do this by preventing `scroll` events that happen within the menu from bubbling up to the document.  Then also listening for `document:scroll` events inside of the component, and closing the menu if it's open.

pjsier:
Ahh that makes more sense, thanks! I'll close this then, and it sounds like you have a better sense of how it should work if you want to take it?

Lane:
Yep sure, I will take this one.

PR #370, opened 2018-01-05
Show loading indicator on data, not render

pjsier:
Potential fix for #336. You can try out the prototype here http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/dataloading-indicator

Instead of looking for whether the full map has loaded, it's now just looking at whether source data is loaded. It doesn't totally resolve the issue of loading showing up on hover, but it seems much more rare from the testing I've done. I was never able to get the flickering issue, but this could potentially resolve that as well


Lane:
Nice, can't replicate the flicker on the prototype link, so that's great! Still get it on hover sometimes, but I think that's less of an issue.

PR #377, opened 2018-01-07
Min export year is min line year

### LINKED ISSUE #376: Should export minimum year default to line minimum year?
Currently the minimum export year defaults to the currently selected year on the map/bar graph, but this means for PowerPoints that the line graph starts later than I would think people expect. Should we set the default minimum year to the default minimum line graph year and base it off of the line graph years instead of the currently selected year? Making a PR for this since it's quick, but we can close it if we decide against it

pjsier:
Closes #376 if we decide to go that way. It seems more intuitive to use the line graph defaults, otherwise if someone selects a feature on the map and exports to a PowerPoint, the default line chart is only for 2016 and 2017


Lane:
Yep, makes sense to me.

PR #378, opened 2018-01-08
Add linting, include in Travis script

pjsier:
Not top priority, but figured it makes sense to clean up some linting errors and add that to Travis for now, since if anything breaks linting that's crucial we can just disable the rule


Lane:
Ahhhh, just realized I forgot to turn on linting on my new laptop!  Thanks for cleaning things up.

There was a conflict with adding `HostBinding` to `map-tool.component` for the `software-button` class.  I moved the `HostBinding` to the `app.component` because we will likely need to use it for the ranking tool.  If this is good with you, then this PR is good to merge in.

PR #390, opened 2018-01-08
prevent propagation from select scrolling

### LINKED ISSUE #188: UI Select: Turn off page scrolling when menu is open

Lane:
Closes #188, does not scroll page if user is scrolling a dropdown list.  If they scroll outside of the dropdown list, the list closes and the page scrolls.

select-scroll


pjsier:
This looks great! One separate issue it might bring up is that the `overflow: hidden` property on map tool doesn't actually stop scrolling.

PR #398, opened 2018-01-09
Drop parallax for fixed position

### LINKED ISSUE #229: Change map to use CSS parallax on scroll
https://keithclark.co.uk/articles/pure-css-parallax-websites/

Calculating the scroll position on the asynchronous thread leads to jumping and is bad for performance.

Lane:
Made a few attempts to cut the jank from the parallax scroll of the map and fix #229 

**Attempt 1: Make parallax animation with CSS only (as per: https://keithclark.co.uk/articles/pure-css-parallax-websites/)**

Requires changing the DOM structure pretty significantly, and doesn't easily allow variable heights on the DOM elements.  Decided this was too difficult to implement for the benefit it would provide.

**Attempt 2: Optimize parallax scrolling by setting an update interval, using `requestAnimationFrame`, rounding values, setting position to fixed (as per: https://medium.com/@dhg/parallax-done-right-82ced812e61c)**

This produces about a ~8-10 FPS improvement when scrolling, but it's still janky.  The map jumps to try and catch up with the scroll position. (preview: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/parallax-2/)

### Proposed solution: Drop parallax, used `fixed` position to give layered effect.

This is the only jank-free solution I could produce.  Even just running mapboxgl on the page cuts my frame rate to ~40FPS, so adding any transforms on top of that seems to be a recipe for jank.  I think this is a good compromise between design and performance. (preview: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/parallax-3/)

I have the parallax code in this PR still, but the interval is commented out so it doesn't execute.  We could just uncomment if we decide to turn it on.

### DIFF
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,4 +1,4 @@
-<!-- <app-progress-bar *ngIf="loader.isLoading"></app-progress-bar> -->
+<app-progress-bar *ngIf="loader.isLoading"></app-progress-bar>
 <app-header-bar
   [languageOptions]="dataService.languageOptions"
   [selectedLanguage]="dataService.selectedLanguage"
--- a/src/app/app.component.html
+++ b/src/app/app.component.html
@@ -1,4 +1,4 @@
-<app-progress-bar *ngIf="loader.isLoading"></app-progress-bar>
+<!-- <app-progress-bar *ngIf="loader.isLoading"></app-progress-bar> -->
 <app-header-bar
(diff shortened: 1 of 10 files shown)


pjsier:
I think this makes sense! I've been looking more into performance lately, and the map definitely slows things down already, so simplifying seems like a good idea to me

PR #399, opened 2018-01-09
Improve map event performance

pjsier:
This is a general performance improvement pulling from some of the ideas here https://medium.com/paramsingh-66174/catalysing-your-angular-4-app-performance-9211979075f6

The map hover and zoom events slow things down significantly, and for the most part the consistent events aren't actually needed in Angular. I debounced the hover state updates for both the tooltip and hover outline updates since it's a major draw on the memory. The catch is that now the tooltip doesn't follow the cursor, but unless you're using Chrome at top performance it lags now anyway.

Try it out locally and let me know what you think, but I'm seeing improvements from 3-5 fps to 20-30 while moving the mouse around and panning on the map on Firefox


Lane:
Definitely seeing an improvement here!  My Firefox still chokes on hovering and panning (dropping to less than 1FPS), but it is taking less time to recover.

Issue #189, opened 2017-11-24
Zooming quickly on the map in Firefox moves into the data panel


pjsier:
Using Firefox for MacOS, if you're zooming on the map with a card selected and you swipe down quickly on a track pad to zoom out, it can scroll down to the data panel without selecting "View More Data". Couldn't replicate this on Chrome or Safari.

Relating to this and #163, I'm wondering if there's another option besides `ngx-page-scroll` to use for the scrolling functionality. Their [README](https://github.com/Nolanus/ngx-page-scroll) makes it look like they don't test against Firefox, and if we want to support Firefox it could be worth finding another package

![scroll](https://user-images.githubusercontent.com/8291663/33214490-2e30047a-d0fa-11e7-913d-f86e5d89381d.gif)

Lane:
Good catch!  It looks like this might have to do with the wheel / scroll event being sent to the page instead of the map absorbing all of those events.

I'm cool with switching from `ngx-page-scroll`, but it is only used to handle manually scrolling between the map and the data panel.  I don't think it has any impact on regular scrolling of the page.

pjsier:
From playing around with this more, I think it's most likely just relating to lag. Firefox responds more slowly, so if the map doesn't respond to the scroll event quickly enough the browser seems to default to a normal scroll. I'm good with considering this a won't fix if that's the case

pjsier:
This actually may have to do with using the custom `debounce` function for listening to `wheel` events. I thought the zoom performance issues were mostly related to firing events on all zooms, not just `zoomend`, but looking into the performance stats there's a big toll from calling `setTimeout` multiple times while zooming. I'm gonna try and modify it to see if that improves things

<img width="493" alt="screen shot 2018-01-10 at 7 19 10 am" src="https://user-images.githubusercontent.com/8291663/34774889-c86584ec-f5d6-11e7-9f67-c888a8576336.png">

pjsier:
It seems like we've narrowed this down to a general performance issue that we've been addressing, so I'm closing this in favor of other performance-related issues since we may never fully close it

Lane:
Had this happen to me a couple times today on the staging URL, so I'm going to reopen.  No for sure way to reproduce yet, but seems to happen when I'm zooming from a high zoom level to a low one.

pjsier:
[...]

PR #403, opened 2018-01-10
Use real decorator for debounce

pjsier:
Addresses #189 to the extent that we can probably fix it. I'm getting significantly better performance using a custom method decorator for debouncing wheel events, and from looking into some of the performance issues it seems like that's been a pain point


pjsier:
Since we were only on `@angular/cli` 1.5.4, it made sense to downgrade to 1.5.0 since this is a pretty big benefit

PR #416, opened 2018-01-11
Use display instead of ngIf for data panel

### LINKED ISSUE #410: Pre-render graph and data panel before locations added
Currently, when a user selects the first location the entire data panel has to be rendered because we're using `ngIf`. We should pre-render this and then hide or show it based on whether there are any active features

pjsier:
Closes #410. It's a little alarming how much of a difference it makes. It's not always super visible because we're talking about fractions of a second, but here's a waterfall chart on Firefox for clicking on the first feature currently (with `ngIf`):

screen shot 2018-01-11 at 9 56 11 am

Here it is after this change to `style.display`:

screen shot 2018-01-11 at 9 55 31 am

The click event itself takes a bit less time, there's significantly less time spent on updating styles, and in general the worst case scenario for performance isn't as bad in this case.

### DIFF
--- a/src/app/map-tool/map-tool.component.html
+++ b/src/app/map-tool/map-tool.component.html
@@ -49,7 +49,7 @@
   <div #divider></div>
   <app-data-panel
       id="data-panel"
-      *ngIf="dataService.activeFeatures.length > 0"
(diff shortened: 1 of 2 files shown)


Lane:
Awesome!

PR #426, opened 2018-01-12
Include select fix for touchmove

### LINKED ISSUE #188: UI Select: Turn off page scrolling when menu is open

pjsier:
Fixes #188. I had to add `touchStartY` because `touchmove` events don't have the `deltaY` property


Lane:
I'm still seeing the page scroll once I hit the bottom of the menu with this... I have no idea why though, the `preventDefault` on the touchmove event should fix it.

pjsier:
It might be getting triggered by `touchstart`? Testing that now

PR #429, opened 2018-01-12
Add padding in legend, y axis label

### LINKED ISSUE #425: Graph style adjustments
From Noele: Please add a bit more padding between "Eviction Rate" and the left side of the graph; and then more padding between the numbers and the x-axis/y-axis.

Also need to add padding between the Kansas, Iowa, and Nevada section on the right so they don't all flow together - 20-30px padding should be sufficient.

image

pjsier:
Closes #425. Adds space in the legend labels and axis labels. Dependent on https://github.com/EvictionLab/angular-d3-graph/pull/11 for setting `tickPadding` for each axis

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -866,9 +866,9 @@
       "integrity": "sha1-SlKCrBZHKek2Gbz9OtFR+BfOkfU="
     },
     "angular-d3-graph": {
-      "version": "2.0.9",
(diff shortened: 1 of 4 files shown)


Lane:
EvictionLab/angular-d3-graph#11 published in version `2.0.10`

Issue #437, opened 2018-01-15
Separate S3, CloudFront setup for production


pjsier:

pjsier:
Confirmed working after the latest master merge

PR #448, opened 2018-01-16
Update credentials for prod

pjsier:
Finishes #437. It looks like we needed to switch the region and credentials but it's working fine on a test branch now. Once we merge this to master it should deploy as expected


pjsier:
Closing after fixing the IAM role

PR #452, opened 2018-01-16
Connect bar year and general map year

### LINKED ISSUE #439: Bar chart year selector should match general year selector
Waiting for final confirmation, but it sounds like there's a consensus that the bar graph year selector feels disconnected from the rest of the page. Updating the bar selector year should update the map and general data panel year

pjsier:
Closes #439. This should be set, but is WIP until we get final confirmation on updating this (sounds like everyone is on the same page though).

Removes `barYear` on the data panel component, and updates the general data service year instead. Also sets the `ui-slider` component to update its position when the year input is changed, and doesn't call `updateHighlightFeatures` when the year is updated because it can run into errors when the tile layer switches.

### DIFF
--- a/src/app/map-tool/data-panel/data-panel.component.html
+++ b/src/app/map-tool/data-panel/data-panel.component.html
@@ -70,9 +70,9 @@
               <span aria-hidden="true">&times;</span>
             </button>
             <p class="legend-data">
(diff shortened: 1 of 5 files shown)


pjsier:
Just confirmed that it should work this way, so this is good to merge

Issue #442, opened 2018-01-16
Keyboard nav not working on `ui-select`


Lane:
When a `ui-select` has focus the enter/space/down/up/esc key are not behaving as expected.  (e.g. pressing space scrolls page)

Maybe defaults need to be prevented?

Issue #414, opened 2018-01-11
Style :active/:focus states for tab enabled elements


Lane:
Currently has the default blue outline / dotted border when elements are focused.  Add some styling to so they fit in with the rest of the design.  Noele will provide styles.

**Icons**
#FFE8DE square lay behind any icons

![image](https://user-images.githubusercontent.com/21034/34804118-36175f94-f634-11e7-80ca-0149fe52275e.png)

**Buttons**
TBD

**Time Slider Marker**
TBD

**Inputs**
TBD

PR #457, opened 2018-01-16
Map tool keyboard navigation

Lane:
Sets proper tab indexes and ensures that all interactive elements are operable w/ keyboard only

### DIFF
--- a/src/app/header-bar/header-bar.component.html
+++ b/src/app/header-bar/header-bar.component.html
@@ -1,5 +1,25 @@
 <div class="header-wrapper">
   <div class="header-content">
+    <div class="header-logo">
+      <a href="/">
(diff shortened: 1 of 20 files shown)


Lane:
Should be ready to merge.  Lots of shifting around to ensure elements get focus in the desired order.  Also:

- Fixes #442 by giving focus to the inner button instead of to the component host
- Forwards focus to the map canvas from `app-mapbox` so that it can be navigated w/ keyboard
- Progress on #414, adds some defaults for all tabbable components
- Adds keyboard bindings to `app-ui-switch` component

Lane:
Wait, not ready to merge, I think tests are failing.  Also there will probably be conflicts with the other PRs, so let me merge those in and resolve any conflicts.

Lane:
Wow, no conflicts.  Looks like tests aren't completing because TravisCI is having issues... So, this should be ready to merge.

Issue #455, opened 2018-01-16
Update iOS Safari padding


pjsier:
Safari ends up being a lot more difficult to check for than Android, especially relating to WebViews which don't have the same height issues. The tentative solution is to remove it from the software button check and bump up the scroll indicator just a bit so that it might show through

PR #458, opened 2018-01-16
Update mobile Safari, attribution

### LINKED ISSUE #443: Move attribution to bottom left
The attribution is currently running into the location button on mobile, so we might as well move it to the bottom left instead of bottom right. Here's an example of how to do that  https://www.mapbox.com/mapbox-gl-js/example/attribution-position/

pjsier:
Fixes #443, #455. It got a bit complicated, and I realized midway through that if Safari doesn't have padding the year selector isn't visible unless you select a feature. I'm at the point where I think we'll just have to accept that, since the alternative is shrinking the map significantly. Here's the prototype, http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/device-tweaks

It works overall testing on BrowserStack and my iPhone, but let me know if it looks like I'm missing anything here

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -47,11 +47,9 @@ export class AppComponent implements OnInit {
     this.translate.setDefaultLang('en');
     this.translate.use('en');
     this.onWindowResize();
-    // Add software-button class if browser is Android or iOS Safari
+    // Add software-button class if browser is Android
     const userAgent = navigator.userAgent.toLowerCase();
-    this.softwareButton = userAgent.indexOf('android') > -1 ||
-      ((userAgent.indexOf('iphone') > -1 || userAgent.indexOf('ipad') > -1) &&
-       userAgent.indexOf('crios') === -1);
+    this.softwareButton = userAgent.indexOf('android') > -1;
   }
 
   /** Fired when a route is activated */
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -50,7 +50,7 @@ export class AppComponent implements OnInit {
     this.onWindowResize();
     // Add user agent-specific classes
     const userAgent = navigator.userAgent.toLowerCase();
-    this.iosSafari = ((userAgent.includes('iphone') || userAgent.indexOf('ipad') > -1) &&
+    this.iosSafari = ((userAgent.includes('iphone') || userAgent.includes('ipad')) &&
       (!userAgent.includes('crios') && !userAgent.includes('fxios')));
     this.android = userAgent.includes('android') && !userAgent.includes('firefox');
   }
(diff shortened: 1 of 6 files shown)


pjsier:
Second guessing this a bit. Some of the changes are helpful, but Safari is almost unusable on mobile without the software button padding, and I think it’s more acceptable for links from web view to be a bit more cramped than Safari to be missing content

Lane:
Looks good on Android / iOS chrome, though I think we should find solutions for these:

- Android mobile Firefox:  Has quite a bit of additional padding, leaving not much map space.
- iOS Safari: bottom bar covering up slider or scroll hint when card is selected (as you mentioned)

Looking at the CSS it's a bit tough to determine which height adjustments are being made for what.  For readability sakes, we might want to add a `$softButtonsPadding` or something like that instead of the arbitrary `grid(x)` values in the height calculations.

pjsier:
Updated this so that the attribution is under the location button on mobile and tablet, and on the bottom left on desktop. Also simplified the rules a bit and tried to make them more clear. I added Firefox to the exclusions for `software-button` (not sure if that's the right user agent string for Android though) and added Safari back as well

pjsier:
This last update works great for iOS Safari, but on BrowserStack it looks like a little more than needed for Android. Let me know if it looks that way on a real Android, and if it does I think it makes sense to add a `mobile-safari` class. Firefox seems easy enough to detect from user agent alone

Lane:
Looks good on Android Firefox, but now Android Chrome has the extra padding.

I think it'd make sense to add the following classes to `app.component`:

  - `.android`
  - `.ios`
  - `.chrome`
  - `.ie`
  - `.edge`
  - `.firefox`
  - `.safari`

Using those classes in combination with `.gt-mobile`, `.gt-tablet`, etc. should give us everything we need to add styling overrides for these edge case browser specific things.  Does that make sense, or is that overkill?

pjsier:
I think it feels a little like overkill? It looks like there are two types of padding: general Android software button padding and iOS Safari. It looks like Safari needs more padding than Android, so maybe we can change `.software-button` to `.ios-safari`, and then add `.android` with less padding. Does that work to you? I've also been going off of the BrowserStack heights, so not sure how accurate those are

Lane:
Sure, works for me.  Just want to make sure we aren't creating a new class on `app.component` for every edge case we encounter.

pjsier:
This last update should have the right height for Android. The last thing I want to check is if we can check if iOS is using `WebView`, in which case it wouldn't get the `ios-safari` class applied. It looks like there might not be reliable ways of doing that, but I'm gonna keep looking

Lane:
I am pretty sure it is not possible to differentiate the WebView / Safari with javascript, though I will mail you a beer if you are able to find a way!  Tests are passing locally, so I'm going to merge this.

PR #465, opened 2018-01-17
Base hover outline color on selected features

### LINKED ISSUE #462: Update map hover colors
Noele has mentioned this a while ago, but I don't think we captured it anywhere.  Set the geography outline color on hover to the following:

- When 0 cards are active: #e24000
- When 1 card is active: #434878
- When 2 or more cards are active: #2c897f

pjsier:
Closes #462. The only concern I have about this is that you can't see the blue hover outline in areas that have the darker choropleth shade. I don't think this is that big of an issue though, since we still have the hover tooltip

hover-colors

### DIFF
--- a/src/app/map-tool/map/map/map.component.html
+++ b/src/app/map-tool/map/map/map.component.html
@@ -51,6 +51,7 @@
     [mapConfig]="mapConfig" 
     [eventLayers]="mapEventLayers"
     [selectedLayer]="selectedLayer"
+    [featureCount]="activeFeatures.length"
     (ready)="onMapReady($event)"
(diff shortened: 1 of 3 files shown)


Lane:
Is it possible to have the same style for hover as when a location is clicked, with the white outline?

pjsier:
Yeah, that's actually pretty simple now that you mention it. I'll fix that now

pjsier:
Screenshot with the update

screen shot 2018-01-17 at 11 19 27 am

Lane:
Nice, looks good to me!

PR #476, opened 2018-01-18
Set bar graph x tick format

### LINKED ISSUE #475: Year formatting incorrect on line graph transition
I think we partially fixed this before, but the X axis labels on the line graph are briefly formatted with commas while transitioning from the bar chart.

year-format

pjsier:
Fixes #475. It looks like the transition is using the first axis' `tickFormat`, but since we're not using X axis labels on the bar graph anyway it doesn't seem to hurt to update the format to prevent the weird transition behavior.

year-format-fix


Lane:
Good call, this bug keeps creeping in so this should hopefully fix it for good.

PR #479, opened 2018-01-18
Debounce more wheel/scroll events

pjsier:
General performance update. It seems to make a difference on Firefox at least, and we don't need to be calling either of these functions all the time anyway.

Here's Firefox before the change:
screen shot 2018-01-18 at 11 22 06 am

Afterwards the events happen less often and don't seem to impact performance nearly as much:
screen shot 2018-01-18 at 11 25 09 am


Lane:
Agree that there is definitely a performance issue here, good catch.

The reason I don't have a debounce on the wheel event is that we want it to fire immediately when the wheel scrolling starts, because it is used to prevent an overscroll which causes the map to zoom when the user flicks he trackpad to scroll to the top.  Seen in the screenshot below:

This PR (with debounce):
overscroll

Without debounce:
overscroll-no-debounce

So, I think this needs to be looked into a bit more to find a way to increase performance without having the overscroll zoom the map once it hits the top of the page.

pjsier:
Ah good catch! I might try looking into the RxJS docs for this

Lane:
Yeah, there has got to be something with RxJS that handles this kind of thing.  Also, looks like lodash's debounce function allows for a leading event to be fired, which should do the trick as well: https://stackoverflow.com/questions/34552038/debounce-function-that-fires-first-then-debounces-subsequent-actions

pjsier:
Let me know what you think of this latest update. I switched it to use Observables instead of host listeners, and I use `take` to pull the first 10 events (to be safe) and then debounce after that

Lane:
Still getting the map zoom once I hit the top of the page when two finger scrolling with the track pad.
overscroll2

pjsier:
Ok I misunderstood the original issue, but I think there's a way of fixing this

Lane:
This is real close!  The scroll doesn't happen when scrolling to the top anymore, but if I scroll down with my cursor over the cards, the map will zoom out.

scroll-zoom

pjsier:
This update seems to fix that for me. I made the debounce and throttle times really short, which still gives us the performance boost but doesn't seem to have the same side effects

Lane:
Still seeing it with the latest commit.

Lane:
Sorry, didn't see the update on this one.  The initial zoom doesn't happen when scrolling down anymore, but now we're back to the earlier issue of the map zooming in when scrolling back up to the map with the trackpad.

wheel-zoom

pjsier:
At this point I don't know if there's any way to get the right balance of improving performance on the scroll and also not adding weird scroll behavior. I'm fine with closing this if there isn't another method you can think of

Lane:
Just added a commit that should fix things up, going to do a quick performance check.

Issue #483, opened 2018-01-18
Selecting location in search throws error


pjsier:
Selecting an item from the location search does nothing and throws the error `this.mapComponent is undefined`

pjsier:
This may just be a build error, because this isn't an issue in my local dev branch. Rebuilding the last `development` Travis build (which seemed to have some deployment issues) and will close this if that fixes the issue

pjsier:
Update: Travis is having issues, but I built the site locally with `ng build --prod` and I'm having the same issue. Not sure what's different between the production build and the dev server though

PR #484, opened 2018-01-18
App Menu

### LINKED ISSUE #474: Mobile menus broken
The Map submenu no longer opens, and tapping Search pushes the menu icons down, and puts the search nav above them

The Map submenu opens at a mobile screen size on desktop, but not on mobile devices, which makes me think it has something to do with touch events. The Map submenu bug doesn't affect `master` currently, but the Search position bug does.

Map bug was introduced some time between bbc6524 and 29c79aa, and Search bug was introduced between de6c1b4 and 6c886c7

Lane:
Adds the fly out menu.  Also fixes #474 and updates GT Eesti to the proper font.

menu-bar


Lane:
Wait on this, not ready yet.  Realized that the map button issue on #474 is caused by the same issue with #483

pjsier:
Sounds good! Yeah, I had a feeling, trying it out locally with `ng build --prod` now

Lane:
Okay, latest commit on this PR should fix #474 and #483

PR #487, opened 2018-01-19
Split out some configuration between environments

### LINKED ISSUE #486: Use environments for managing configuration
There are a few API keys and URLs that it's probably easier to manage in the environment files, and it also may be beneficial to build a `staging` environment off of `development` instead of using production for both

pjsier:
Closes #486. Trying to move a lot of the external service configuration to environment variables, especially for the exports link (which we have a dev URL for) and the Mapbox API key. Also updates the build step in Travis with this. A lot of it is redundant, but I figure we can add onto this as we go along, and it could make these changes easier in the future

### DIFF
--- a/.angular-cli.json
+++ b/.angular-cli.json
@@ -29,6 +29,7 @@
       "environmentSource": "environments/environment.ts",
       "environments": {
         "dev": "environments/environment.ts",
+        "staging": "environments/environment.staging.ts",
         "prod": "environments/environment.prod.ts"
       }
     }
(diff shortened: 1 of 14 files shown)


Lane:
This is great! Assuming those tests failed because of the `@angular/dev-kit` issue?  If so, is it fine to merge this?

pjsier:
Yep, just fixed one test that was relying on the wrong endpoint and now this should be good to merge. The tests pass locally and so does the build, so there shouldn't be any issues

Issue #488, opened 2018-01-19
Use OnPush change detection where possible


pjsier:
I've seen `OnPush` change detection as a way of improving Angular performance in a lot of places, and from trying it out on a few components locally it doesn't seem to cause any issues at first glance. Is this something to consider implementing for performance improvements?

Lane:
I've never used it so I'd have to look into it a bit, but if it gives us a measurable performance boost and doesn't introduce any issue then I don't see why not!

pjsier:
I would consider this a low priority or potentially not worth it at all since I didn't see a noticeable performance boost, but it has the potential to cause a lot of complications

PR #489, opened 2018-01-19
WIP: Using OnPush change detection

pjsier:
Looking into #488. I'm not 100% sure of this, so not looking to merge yet, but (other than rankings) the only component where adding this seems to cause issues is `ui-toggle`. Otherwise it looks like we're generally updating full inputs rather than modifying properties so it doesn't seem like a huge shift.

I'm not totally sure about the performance improvement either from local testing, but this is recommended in every article I've seen about Angular performance. Take a look when you have a chance


pjsier:
Closing this for now, since it's straightforward enough to implement down the line, but there are a good amount of unknowns with it and not many clear performance improvements

PR #496, opened 2018-01-19
Throttle year slider events

### LINKED ISSUE #495: Year Slider Lags
Year slider lags when dragging.  Need to determine the expensive function calls and debounce those.

slider-lag

pjsier:
Closes #495. Still not quite at the same speed, but with debounce it just sits there until you're done scrolling, and increasing the throttle too much makes the scroll a bit shaky. This was the best balance I could find, and we'll at least get the general performance boost

year-slider-drag


Lane:
I don't think it's the actual event that needs to be debounced / throttled.  It's what happens when the value changes that is causing the delay.  I am getting similar delay with the throttling.  I have a branch with the handler debounced and it seems to fix this.

pjsier:
Sounds good! I'll close this

Issue #503, opened 2018-01-22
Add hover tooltip back to cities, fix removed text layers


pjsier:
It looks like the city hover tooltip was removed in the map style update. I've also seen a few errors looking for layers like `tracts_text` which was removed in that update as well, so there are likely some places in the code statically looking for `${layer}_text` that will need to be updated

PR #507, opened 2018-01-22
Update city labels, null check on tooltip

### LINKED ISSUE #502: Use us-cities-10 source for city labels
City labels are currently being sourced from OpenMapTiles, but that's leading to some weird inconsistencies where mid-sized cities in their data aren't represented in our census information:

screen shot 2018-01-22 at 8 40 39 am

We should be able to switch to using the `us-cities-10` layer and filtering based off of the population as a rough proxy so that labels line up

pjsier:
Closes #502 and #503. Prototype of the changes here http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/city-labels/

### DIFF
--- a/src/app/map-tool/map/map.service.ts
+++ b/src/app/map-tool/map/map.service.ts
@@ -329,29 +329,38 @@ export class MapService {
     const features = this.map.queryRenderedFeatures(e.point, { layers: layerIds });
     if (features.length) {
       const feat: MapFeature = features[0];
-      const labelFeatures = this.map.queryRenderedFeatures(undefined, {
-        layers: [`${feat['layer']['id']}_text`],
-        filter: [
-          'all',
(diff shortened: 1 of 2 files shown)


pjsier:
Null check is fixed so this is ready for review

Lane:
Not sure if this is a big deal, but I've noticed some cities also have the county names in them on our `us-cities` source.

image

pjsier:
Yeah, I noticed that too. It's a weird quirk of the census data that some census places are both counties and "cities". I'm good with leaving that as-is, since these labels are more for general orientation, but we could probably filter out the ones that end with county if you think it's confusing

Lane:
It doesn't really bother me, will merge in for now and get some feedback from the team if it's okay as is.

PR #519, opened 2018-01-24
Additional tests

pjsier:
Adding some more tests on the map component and service


Lane:
Nice, this is much appreciated! I've been neglecting writing tests out of concerns of missing deadlines, but will start addressing that as well now that we have more time.

PR #521, opened 2018-01-24
Fix slider jump

### LINKED ISSUE #518: Slider jumping
Slider sometimes emits an incorrect value, causing the marker to "jump"

image

Lane:
Closes #518 by emitting change events instead of modifying the value directly.

Also calls to `updatePosition` are ignored if the scrubber is currently pressed.


Lane:
Also ditches `this.cdRef.detectChanges();` because we're not modifying the value internally in the component anymore.

Lane:
Scratch that lat comment... `this.position` is still changed in `ngAfterViewInit` so we need to flag a change detection with `this.cdRef.detectChanges();`

Issue #432, opened 2018-01-12
Update eviction stats in data panel cards


Lane:
First two stats on data panel cards are "Eviction Rate" and "Evictions", but should be "Eviction Rate" and "Evictions Per Day".  Eviction rate should also show a comparison to US average.

![image](https://user-images.githubusercontent.com/21034/34898030-3934e752-f7ae-11e7-9b73-d08a03dbddd2.png)

pjsier:
Just added a US average data creation step, so we can load it similarly to the rankings. It's at https://s3.amazonaws.com/eviction-lab-data/avg/us.json, and it's just the eviction properties by year

PR #522, opened 2018-01-24
Add US average to graph, card data

### LINKED ISSUE #463: Add US average to graphs
Not sure what was settled on, but adding the US average as a point of comparison has come up a few times

pjsier:
Closes #463. Splits `locations` on the data panel into `displayLocations` for anything that should have its own location card and `allLocations` for anything that should be displayed on the graph. Updates `$color4` in sass to be the color mentioned in the issue (it was a dark gray). Also fixes an unrelated typo in the data service (`bbox` was being checked twice when it should be checking for `bbox` and `layerId` according to a comment)

us-avg

### DIFF
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -302,6 +302,16 @@ export class DataService {
       +feature.properties['east'],
       +feature.properties['north']
     ];
+    // Add evictions-per-day property
+    Object.keys(feature.properties)
+      .filter(p => p.startsWith('e-'))
+      .forEach(p => {
+        const evictions = +feature.properties[p];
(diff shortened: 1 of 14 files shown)


pjsier:
Updated it to also close #432. Generates the evictions per day property in the data service (which required updating the `MapFeature` interface and specifying type in some functions), fixes the order of properties displayed in the cards, and shows a comparison to the US average eviction rate if the location has data available.

I'm not clear on the best way to position the text without messing up the justify content display, so I just set the comparison as `position: absolute` and and that seems to be working

screen shot 2018-01-24 at 3 47 00 pm

PR #523, opened 2018-01-25
Enhancement/route graph type

### LINKED ISSUE #451: Include graph type as URL parameter
I remember someone mentioning adding the graph type to the URL as a query parameter, it seems straightforward enough to me

Lane:
Adds `graph` as query parameter. Closes #451 

Also re-organizes attributes in `data-panel.component` so they are grouped by their relevant section.


pjsier:
The tests pass, but there was a linting error that I just fixed. Otherwise this looks great! Thanks for cleaning up the attributes in the data panel

PR #539, opened 2018-01-26
Selected item focus in dropdowns

### LINKED ISSUE #535: Set focus to proper item in `ui-select`
When opening a `ui-select` focus is given to the first element in the list, but it should focus the currently selected item.

Lane:
Gives the selected dropdown item focus when the menu opens, closes #535


src/app/ui/ui-select/ui-select.component.ts
```diff
@@ -72,6 +72,9 @@ export class UiSelectComponent implements OnInit {
     if (this.dropdownList) {
       this.dropdownList.nativeElement.scrollTop = 0;
       this.setScrollMax();
+      if (this.open) {
```
pjsier: It looks like this is left over from something?

Lane: whooops

Issue #500, opened 2018-01-22
Update hover debouncing


pjsier:
We may want to relax the debouncing on map hover events, in particular `mouseleave`, because it didn't seem to be the main cause of performance issues when we originally changed it and it currently looks a bit laggy

PR #540, opened 2018-01-26
Refactor hover popup and outline

### LINKED ISSUE #537: Update tooltip immediately/quickly, debounce hover outline
It looks like the hover outline update takes significantly longer than the tooltip update, so we should try to prioritize the tooltip updating quickly since it's helpful for quickly scanning the map and debounce the outline update

pjsier:
Closes #537 and #500.

* Moves all hover interactions into the Mapbox component running outside of Angular (rather than splitting them in between the map service and component)
* Splits the popup (renamed from tooltip for consistency) updates into location and content. The location update doesn't result in any performance hit since it's a cheap calculation, and even if the text inside it doesn't update until a debounced feature change it looks a lot smoother
* Simplifies the conditional on whether a tooltip should show up. Before we were hitting the map multiple times, now since the way we show labels has changed it's a lot more straightforward
* Updates `getUnionFeature` to only do any calculations if the feature is missing area compared to its bounding box, or if it is using the bounding box geography in which case it should be updated. This is one of the worst functions for performance, but we don't have another way of pulling the full data with vector tiles, so minimizing how much it gets called helps
* Removes `states_text` since we're not using it
* Fixes references to `feature.properties.bbox` since the standard way of using it, included in the interface, is `feature.bbox`
* Adds tests for the updated geometry behavior

After this, I can hover around at this speed with counties on the default view and still average around 35fps on Firefox

hover-pr

### DIFF
--- a/src/app/data/data-levels.ts
+++ b/src/app/data/data-levels.ts
@@ -9,7 +9,6 @@ export const DataLevels: Array<MapLayerGroup> = [
       'states',
       'states_stroke',
       'states_bubbles',
-      'states_text',
       'states_null'
     ],
     'minzoom': 0,
--- a/src/app/data/data.service.ts
+++ b/src/app/data/data.service.ts
@@ -204,7 +204,7 @@ export class DataService {
       .find(f => f.properties.GEOID === feature.properties.GEOID);
     if (exists) { return null; }
     // Process feature if bbox and layerId not included based on current data level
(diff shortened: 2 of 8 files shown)


Lane:
Daaaaang this is nice! 💯  Feels way better hovering on things in Ubuntu / Firefox as well, which is where it was worst for me.

PR #543, opened 2018-01-27
Don't setup hover interactions on mobile

### LINKED ISSUE #541: Hover interactions still happening on mobile
#540 still left some of the observables and mouse events on mobile, but they need to be removed

pjsier:
Fixes #541, should also give us some performance benefits on mobile without some of these events being fired in the background


pjsier:
Also removes the extra white outline for features that can't be selected on hover

screen shot 2018-01-27 at 10 21 02 am

pjsier:
Also just removed the `debounceTime` on the call to `updatePopupContent` since it didn't actually improve performance since `distinctUntilChanged` is used and the function itself isn't calling any of the more resource-intensive map functions anymore

Issue #544, opened 2018-01-27
Disable scroll on UI slider for mobile


pjsier:
Unless you move the slider exactly left and right on mobile, you end up scrolling a bit down the page on mobile. We should disable this for mobile

PR #546, opened 2018-01-29
Disable scroll during zoom

### LINKED ISSUE #189: Zooming quickly on the map in Firefox moves into the data panel

pjsier:
Should close #189 by disabling all scrolling on the `body` element while a map zoom is active. Originally I had just been bubbling the map `zoomstart` and `zoomend` events, but that seemed like a lot of filler. Since this is functionality we may want to use in other parts of the tool for #544, moving it into a separate service seemed to make sense

### DIFF
--- a/src/app/toggle-scroll.service.ts
+++ b/src/app/toggle-scroll.service.ts
@@ -1,0 +1,12 @@
+import { Inject, Injectable } from '@angular/core';
+import { DOCUMENT } from '@angular/platform-browser';
+
+@Injectable()
+export class ToggleScrollService {
+  set allowScroll(val: boolean) {
+    this.document.body.style.overflowY = val ? null : 'hidden';
+  }
+
+  constructor(@Inject(DOCUMENT) private document: any) { }
+
+}
--- a/src/app/toggle-scroll.service.ts
+++ b/src/app/toggle-scroll.service.ts
@@ -3,8 +3,13 @@ import { DOCUMENT } from '@angular/platform-browser';
 
 @Injectable()
 export class ToggleScrollService {
+  /**
+   * Setting position fixed on body will prevent scroll, and setting overflow
+   * to scroll ensures the scrollbar is always visible.
+   */
   set allowScroll(val: boolean) {
-    this.document.body.style.overflowY = val ? null : 'hidden';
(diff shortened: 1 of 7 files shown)


src/app/toggle-scroll.service.ts
```diff
@@ -0,0 +1,12 @@
+import { Inject, Injectable } from '@angular/core';
+import { DOCUMENT } from '@angular/platform-browser';
+
+@Injectable()
+export class ToggleScrollService {
+  set allowScroll(val: boolean) {
+    document.body.style.overflowY = val ? null : 'hidden';
```
Lane: Calling document directly works because it is a global var, but if we're accessing the `document` through injection should this be `this.document.body.style.overflowY`?

pjsier:
Just made that update and rebased the commit

Lane:
I keep getting "No provider for ToggleScrollService" errors when I load this branch locally.

image

Lane:
Alright, good news is that I can't replicate #189 with this change!  Bad news is we have jumpy elements every time overflow is hidden on the body:

elements-jump

pjsier:
Hmm, that's weird, especially since I'm setting `overflow-y` and not `overflow` generally. Any ideas on how to resolve that?

Lane:
We can disable scroll and force the scrollbar to stay visible as one potential fix (see latest commit).  We still have the scrollbar flashing a bit which might not be ideal, but the page width doesn't change this way.  Would want to test what it looks like on Windows because it has the most visible scrollbar treatment.

elements-no-jump

pjsier:
Windows 10 seems to look alright on Edge from a prototype link I deployed:

disable-scroll

Lane:
Cool, thanks for checking that out, merging.

PR #548, opened 2018-01-29
Initial setup for limited embed view

### LINKED ISSUE #509: Embed as iframe share option
We've been thinking of making an option to embed the current map view as an `iframe` so that smaller news outlets and sites could pull a customized view directly into their content (I've seen it done with CARTO and Mapbox maps before).

It could exist as an additional option in the share menu, or it could potentially be a sub-option on the share a link popup. Either way, we'd likely need to highlight it somewhere that's more media-focused since I don't imagine people will be looking for that on their own.

We could trigger the `iframe` view with a URL parameter that limits the view, potentially removing most/all controls for changing the map data since the `iframe` will probably be a much smaller size.

pjsier:
Progress on #509. This keeps getting merge conflicts, so if the scaffolding looks right we might as well merge it as something to jump off of


Lane:
Merging in the scaffolding works for me.  I'm wondering if it'd make more sense to provide this as a separate route that directs to an embed component instead of the map tool component? That could be a refactoring task for later.

PR #552, opened 2018-01-30
Unset IE scroll styles

### LINKED ISSUE #551: IE gets stuck with scroll disabled
It looks like IE requires CSS values to be set to an empty string rather than `null` to remove them. This also works for other browsers, so I'll make the update

pjsier:
Fixes #551. Not sure why, but IE 11 requires an empty string to unset styles (I checked it out in Browserstack). This works across Firefox and Chrome as well, so looks like it's safe generally


Lane:
Good to know, thanks!

PR #554, opened 2018-01-30
Refactor data panel

Lane:
Working on a few adjustments

  - Separate graph area into its own component
  - Consolidate all data attributes into one data source to avoid hard coding property ID's for translations, formatting, etc.


Lane:
Alright this should be good to merge. Apologies for the large amount of changes, had to rework a few things I didn't anticipate.  Changes include:

  - `MapDataAttribute` objects have a `type` property to determine what they are for (current values are `bubble`, choropleth`, `card`).  We now have one array with all data attributes, instead of separate bubble attributes and choropleth attributes.  
  - `MapDataAttribute` can have a `format` property that determine how they're displayed (currently: `percent` or `dollar`)
  - `MapDataAttribute` objects have an `order` property that determines the order it displays in the data panel location cards.
  - Location cards now take `MapDataAttribute[]` as input that determines what the card shows, instead of an object with `key` : `langKey` mappings.  This way, we don't need so many hard coded attribute ID's, and the translation of the attribute labels is already done.
  - Separated the graph area from `data-panel.component` into its own component (`eviction-graphs.component`).
  - Selected values for line graph start / end year, and whether or not US Average is active is stored in `DataService` so that the file download dialog can access them.
  - US Average is always grey in the graph.

pjsier:
Starting to look through this now, and one thing I'm noticing is that the US average staying gray works, but we need to update its tooltip to be the same:

screen shot 2018-01-30 at 5 26 11 pm

Lane:
Ahh yep, good catch.  I'll fix that now.

pjsier:
This looks great! The refactor of attributes in particular makes things a lot more straightforward

PR #560, opened 2018-01-31
Fix slider jump when released

### LINKED ISSUE #518: Slider jumping

Lane:
This adds a transition to the slider so it should animate into place when released.  Also has the added benefit of transitioning the scrubber when using it with the keyboard.

Should fix #518 on iOS Chrome, but I don't have iOS available to me to check right now.


pjsier:
This seems to address it from a prototype I've been testing out

Issue #379, opened 2018-01-08
Implement Help section


pjsier:
I didn't see this captured anywhere, but right now we don't have any functionality for the help section. Not sure where this fits in timeline-wise

Lane:
Still not sure what this looks like, content is still being developed.  Marking as design needed, will fit into timeline once we know what it looks like.

PR #564, opened 2018-02-01
Help Dialog

Lane:
Adds functionality for #379, still waiting on content and to determine what desktop functionality will be (if any).  HTML content can be added in the i18n files when ready.

help-dialog


Lane:
Added in the content from the content matrix, so this should close #379

PR #566, opened 2018-02-02
Map Tool Module Refactor

Lane:
Pull map tool related data, components, and services into the map tool module.  Split data service into `map-tool.service` and `routing.service`.  Filters out default values from the URL. Add scaffolding for embed component.

Still need to do some testing to make sure everything is working as expected.


Lane:
Alright, this should be good now.  Tried to clean things up a bit so that mapping tool specific stuff is in `/src/app/map-tool` and anything in the `/src/app` folder is shared between the mapping tool and rankings.

Here's a more detailed list of what has been done:

  - Move all map data from `/src/app/data` to `/src/app/map-tool/data`
  - Refactor `data.service` into `map-tool.service`, move non-map tool related functions into related components (E.g. header translations into `header.component`)
  - Creating `routing.service` to setup app routes, and to handle pulling data from the route and providing it to the `map-tool.service`
  - Scaffold an embed route and embed component that contains only the map.
  - Create a `services.module` that contains all root level services for the map tool and rankings.
  - Pull `location-cards.component` out from the UI module, and put it in its own module under `map-tool`, as it is only used within components there and not in rankings.
  - Pull `ui-map-legend.component` from the UI module, and put it in `map.module`, as it is only used within the map.

PR #595, opened 2018-02-09
Unfix panel before footer

### LINKED ISSUE #588: Unfix ranking data panel at the bottom of the page
The data panel with the selected location should remain fixed to the bottom of the viewport until the user has scrolled to the bottom of the list.  At that point it should scroll with the page, so it does not overlap the footer.

pjsier:
Closes #588. There are still styling issues with the scale, but ignoring those since it's being removed in #594.

unfix-scroll


src/app/ranking/ranking-tool/ranking-tool.component.ts
```diff
@@ -116,6 +125,15 @@ export class RankingToolComponent implements OnInit {
     }
   }
 
+  /**
+   * Update fixedPanel based on whether the scroll position is at the end of the element
+   */
+  private onScroll() {
+    const content = this.el.nativeElement.querySelector('.page-content');
+    this.fixedPanel = content.getBoundingClientRect().bottom >
```
Lane: A note on performance here, we should probably calculate the position on the page where to fix the element in `ngAfterViewInit` and on window resize so it is not calculated on every call to `onScroll`.  Calls to `getBoundingClientRect()` are pretty costly, see here: http://dcousineau.com/blog/2013/09/03/high-performance-js-tip/

Lane:
One more fix up on this, there is a bit of a "jump" on the panel when it becomes fixed / detached from the viewport.  Ideally it would become detached when the panel hits the exact position it should scroll on the page so there is no jump.

panel-jump

pjsier:
Makes sense, I'll make those changes

pjsier:
It's still a little odd because of the debouncing, but that seems to mostly address it for me

Issue #580, opened 2018-02-07
Ranking Tool Sidebar Component


Lane:
Ranking tool needs sidebar component that contains the location search, and all of the UI elements (`ranking-ui.component`) for manipulating the current view.

**Mobile**
  - Sidebar can be toggled on / off with search & filter button
  - Sidebar fills viewport

**Tablet**
  - Sidebar can be toggled on / off with search & filter button that is fixed to the left side of the screen
  - Sidebar slides in from left and overlaps ranking list

**Desktop**
  - Sidebar displays on the left side of the ranking list
  - Sidebar is sticky as the page scrolls (when vertical space permits)

PR #596, opened 2018-02-10
make ranking-ui a sidebar, rename toggle scroll service, sticky elements

Lane:
Progress on #580, moving `ranking-ui` out from the tab bar and styling as a sidebar.  Still need mobile implementation.

This also adds `position:sticky` to the sidebar and data panel, which might be a better approach for #588.  Support is there for everything except IE11, and I think we can just use `position: fixed` as a fallback for that and set the z-index of the footer higher than the panel so it just slides under.

scroll-panel-rankings


pjsier:
Changing that approach for #588 makes sense, but we probably don't even need `position: sticky` right? Just setting the panel to `position: fixed` and then increasing the footer `z-index` should work, right?

PR #613, opened 2018-02-14
Update query params on rankings

### LINKED ISSUE #610: Add query params to rankings tool
Include selected index and language query params for rankings

pjsier:
Closes #610. Sets `index` and `lang` as query params, and uses the `RoutingService` rather than calling routes directly

### DIFF
--- a/src/app/map-tool/embed/embed.component.ts
+++ b/src/app/map-tool/embed/embed.component.ts
@@ -25,7 +25,7 @@ export class EmbedComponent implements OnInit {
   }
 
   ngOnInit() {
-    this.routing.getMapRouteData().take(1)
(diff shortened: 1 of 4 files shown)


Lane:
Hmm.. I was thinking we'd have the ranking tool reflect changes based on a subscription to the current route, instead of manually pushing route changes like we are doing with the map tool.  The routing on the map tool was kind of an after thought, so it's not put together in the best way. This works though, just a little more complex than it needs to be.

pjsier:
Ahh I was basing this off of how the map tool was set up. I can flip that if it makes things easier

pjsier:
Closing in favor of #615

PR #615, opened 2018-02-14
Alternate routing strategy on rankings

Lane:
@pjsier Here's the location updates based on subscribing to the URL changes.  It makes sense to me to have our routing implementations consistent between map tool / ranking tool, but I kind of wish I had implemented routing in the map tool this way.   If I had, we wouldn't need `routing.service`, we wouldn't need to `updateRoute()` every time something changes, and we wouldn't need to `take(1)` on load.

You've obviously put in a lot of work restructuring the ranking tool routes, and really I'm fine with either approach.  What are your thoughts?

### DIFF
--- a/src/app/ranking/ranking-tool/ranking-tool.component.ts
+++ b/src/app/ranking/ranking-tool/ranking-tool.component.ts
@@ -18,11 +18,29 @@ export class RankingToolComponent implements OnInit {
   /** tab ID for the active tab */
   activeTab = 'evictions';
   /** string representing the region */
-  region = 'United States';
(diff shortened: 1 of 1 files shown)


pjsier:
I like this strategy a lot better, the only thing I can see us needing to add is the `lang` query param, but that still works with this approach. I can add that if it works

Lane:
Sure, that works for me.  Should I open a new issue for the lang query param or would you rather add it in to this PR?

pjsier:
I think it's straightforward enough to add to the PR, I'll make that change

PR #616, opened 2018-02-14
Fix mobile vh scroll jump with delayed transition 

### LINKED ISSUE #542: Mobile vh scroll jump
The scroll jump when the viewport size changes on mobile isn't a huge issue, but it's a bit annoying, so it might be worth implementing some kind of fix like this https://stanko.github.io/mobile-chrome-vh-units-fix/

pjsier:
Potentially fixes #542. I was looking into this issue and found this article: https://abdus.co/blog/preventing-vh-jump-on-mobile-browsers

I'd seen a few JS fixes, but since we're using `vh` in a few different places along with `calc` it seemed like that could get complicated. Option 2 in that article seemed interesting because it's pure CSS, and it actually seems to work great. I have the delay set for an hour, since I don't imagine anyone would have the site open on their phone for that long.

Most of the changes came from having to move the scroll indicator into the map tool rather than the data panel, since we were changing the overall element height to get the scroll indicator to show up which would mean we couldn't delay the transition without breaking that.

I have a prototype here which seems to work well for me on iOS Chrome, Firefox, and Safari as well as Chrome on Android through BrowserStack: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/vh-fix/

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "eviction-maps",
-  "version": "0.0.0",
+  "version": "0.9.0",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -36,7 +36,6 @@ export class AppComponent implements OnInit {
   @HostBinding('class.gt-large-desktop') largerThanLargeDesktop: boolean;
   @HostBinding('class.ios-safari') iosSafari = false;
   @HostBinding('class.android') android = false;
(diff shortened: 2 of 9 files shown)


pjsier:
I'm actually going to need to change how the device specific padding is loaded, that still isn't totally working

Lane:
Cool!  Haven't seen the transition delay hack before.  This is seems to be working fine in Safari / Chrome on iPhone 5 for me.  Is it good to merge?

pjsier:
Yep, it should be ready to go! I just kept running into small bugs each time I thought it was done, but those should be settled

Issue #651, opened 2018-02-19
Instructions, functionality for sharing embed view


pjsier:
The embed view in #509 is basically finished (other than needing a bubble legend which is captured in #568), but we don't currently have any UI or location to explain how it is shared. I would lean toward having it separate from the tool itself, but since it's based on map selections I'm not sure that's possible

pjsier:
Superseded by #675

Issue #568, opened 2018-02-06
Bubble legend included somewhere


pjsier:
I know we had talked about putting this in help, but since it isn't in the current design and since we have some extra time I wanted to raise it here.

We'll probably need to include a bubble legend for the embed component since we won't have the space for location cards, so it might be worth considering adding it to the general map. Maybe something like this, where it's just two bubbles (changing size instead of color) on the left side of the current choropleth legend?

<img width="315" alt="screen shot 2018-02-06 at 7 56 38 am" src="https://user-images.githubusercontent.com/8291663/35863157-62214002-0b13-11e8-8f99-3a49deade260.png">

Here's another option that's really simple and compact, and used in a similar context

<img width="439" alt="screen shot 2018-03-17 at 4 50 03 pm" src="https://user-images.githubusercontent.com/8291663/37560275-420403fe-2a03-11e8-97aa-b4ce9ceda232.png">

pjsier:
One resource on this since it'll have to be updated dynamically https://www.mapbox.com/help/gl-dds-map-tutorial/

pjsier:
@james-minton this has been on the backburner, but it looks like we'll need it

noelelusano:
@pjsier @james-minton working on this; given the amount of stuff in the UI will need some focused time today to iterate on it.

noelelusano:
![image](https://user-images.githubusercontent.com/6036323/37687132-9ddf439a-2c57-11e8-9fbf-e2d9df34862f.png)

• please go by the sketch file as fonts vary between numbers and text.
• spacing is in increments of 8 as much as is possible.
• i've eliminated all letter-spacing on numbers and text here.

sketch:
https://app.box.com/s/go5r2dyjb3uqeu64dhw0qqmhin7bbmin
(under "Symbols - Legends")

pjsier:
Closed by #861

PR #654, opened 2018-02-19
Finish up iframe embed feature

### LINKED ISSUE #509: Embed as iframe share option

pjsier:
Closes #509, since the rest of the functionality is in #568 and #651. Getting pym.js set up was a bit of a pain initially since it sets up query parameters that are swallowed by our hash routing, but after swapping those back in it looks like this will be a lot easier than trying to create an iframe with the right set of parameters

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -8834,6 +8834,11 @@
       "integrity": "sha1-wNWmOycYgArY4esPpSachN1BhF4=",
       "dev": true
     },
+    "pym.js": {
+      "version": "1.3.2",
+      "resolved": "https://registry.npmjs.org/pym.js/-/pym.js-1.3.2.tgz",
+      "integrity": "sha512-/fFzHFB4BTsJQP5id0wzUdYsDT4VPkfAxk+uENBE9/aP6YbvhAEpcuJHdjxqisqfXOCrcz7duTK1fbtF2rz8RQ=="
+    },
     "q": {
       "version": "1.5.1",
       "resolved": "https://registry.npmjs.org/q/-/q-1.5.1.tgz",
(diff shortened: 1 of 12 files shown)


Lane:
Looks good.  For organization purposes, could we keep the embed related CSS in `embed.component.scss`?

Lane:
actually, it's probably fine for the most part, just looking at `app.component.scss` mostly.  Also, not sure I'm understanding what the `html.embedded, html.not-embedded` rules are doing?  Wouldn't that apply to everything?

pjsier:
That was actually going to be my question, moving it out of `app.component.scss` definitely makes sense since I set that up when I was trying to get it to work, but do the styles in `map.component.scss` still work?

And `embedded`/`not-embedded` are both added by pym.js, they're not any of our host-bindings, but `not-embedded` was only there from testing. It's only applied when activate the pym child, but I'll remove it anyway since we don't need it

Lane:
yep, the styles in `map.component` are fine, but would you be able to put the `app.component` stuff in `embed.component`?

pjsier:
Yep! Updated that in the latest commit

Lane:
I think you maybe moved the stuff from `style.scss` instead?  Still seeing the embed related stuff in `app.component.scss`

pjsier:
Whoops, yeah I got them mixed up. I can make that change tonight

pjsier:
Removed everything except the `display: none` on embed branding, let me know if that works!

Issue #624, opened 2018-02-15
Social share general updates


pjsier:
Update social share text across the tool, also make sure social share buttons are included in all the places we planned for them to be (the graph had been mentioned as one additional place

Lane:
I think we're good here, and the remaining updates should be covered by #203

@pjsier feel free to close if you agree

pjsier:
Works for me! Closing

Issue #623, opened 2018-02-15
Rankings social share


pjsier:
Add social share buttons and robo-text generation to rankings page

pjsier:
We may need to add state data to the CSV loaded in the city rankings to generate this the way we want, because aggregating the cities would be significantly different than the state numbers

PR #657, opened 2018-02-20
Social share text updates, rankings social share

pjsier:
Progress on #623, #624, #658 

* Updates the Twitter text generation for the map tool to the latest draft
* Adds a default tweet for when no locations are selected on the map
* Sets up social share for the rankings page
* Robo-text generation for tweets on the rankings page (still need to split out that function depending on the tab)

screen shot 2018-02-23 at 2 55 24 pm

screen shot 2018-02-23 at 2 55 33 pm


pjsier:
The `updateTwitterText` functions could definitely be cleaned up, but I think that makes sense for refactoring down the line

Lane:
Merging in, but if you could break down `updateCitiesTwitterText()` into a few single task functions it'd be much appreciated! Not sure I'm following everything that's happening there.

PR #660, opened 2018-02-20
Unsubscribe from translate on component change

### LINKED ISSUE #659: Changing rankings language redirects to map tool
translate-bug

pjsier:
Closes #659. After looking into it a bit more, the translate redirect only changed the section of the tool if I'd started on a different one. It looks like the initial subscription was never canceled, so it fires the first event which in this case was to update the map tool page. Based on [this StackOverflow post](https://stackoverflow.com/a/41177163) it describes the recommended way to unsubscribe, so I implemented that here and it seems to work.

Not sure how far reaching this is, or if it might be worth adding this to other components that listen to general events, including child components of the map and rankings.


Lane:
Ah yes, good catch.  We'll definitely need to make a pass and make sure we don't have any hanging subscriptions.

PR #678, opened 2018-02-21
Mobile style fixes

### LINKED ISSUE #556: Back to Map button flickers on iOS after clicking map selector
After tapping the map button for the first time, the overlay and "Back to Map" button flicker on the page for a second or two. I think this showed up after #536, but not sure

pjsier:
Some small mobile fix items:

* Fixes a transparent gap between the map and data panel

screen shot 2018-02-21 at 4 26 32 pm
Fixed: 
screen shot 2018-02-21 at 5 25 10 pm

* Makes the year slider marker a little more narrow so that it doesn't overlap the year label

screen shot 2018-02-21 at 5 25 53 pm
screen shot 2018-02-21 at 5 25 03 pm

* Fixes #556, unless you had a different method in mind

### DIFF
--- a/src/app/map-tool/data-panel/data-panel.component.scss
+++ b/src/app/map-tool/data-panel/data-panel.component.scss
@@ -5,17 +5,16 @@
 .data-header {
   @include dataPanelHeader(14px);
   text-align: center;  
-  margin: grid(3) grid(4) 0;
+  padding: grid(3) grid(4) 0;
   p { margin:0; }
-  // margin: grid(5) grid(4);
 }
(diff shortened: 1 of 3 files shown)


Lane:
The fix for #556 is good with me!

The scrubber size and label margins are as provided in Noele's mockups, so I don't know if it's a big deal that it overlaps the right label a bit.  Might want to check with her on that.

Lane:
Noele suggested dropping the scrubber width by 8px and shortening the track a bit to avoid the overlap, it was a small change so I just went ahead and put it in.  Merging.

PR #683, opened 2018-02-22
Update public data link

pjsier:
Updates the public data link and sets it to open in a new tab. I don't feel strongly about opening in a new tab, it's just generally what I've done for external links so let me know if you want to take that out


Lane:
Opening in a new tab makes sense to me!

PR #702, opened 2018-02-24
Disable downloads when no locations selected

### LINKED ISSUE #658: Default state for data panel share items (or hide)
Users can still scroll to the data panel even if they haven't selected any features if they scroll on the top bar. If they try to click on the download or social share icons, they'll get something like this:

screen shot 2018-02-20 at 10 07 34 am

Should we hide this div entirely and only allow people to access the footer below it?

pjsier:
Closes #658 in combination with #657.

disabled-reports

### DIFF
--- a/src/app/map-tool/data-panel/data-panel.component.html
+++ b/src/app/map-tool/data-panel/data-panel.component.html
@@ -104,7 +104,15 @@
           </svg>
           <h2>{{ 'DATA.DOWNLOAD_HEADING' | translate }}</h2>
           <span>{{ 'DATA.DOWNLOAD_HINT' | translate }}</span>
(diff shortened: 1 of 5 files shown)


pjsier:
@james-minton this is the proposed fix for #658

james-minton:
@pjsier Looks great. Feel free to check with Noele to see if she has the bandwidth before she leaves on 3/1 to think about what a disabled button style would look like, or if it's needed given what you've done here. (I think this popover is going to be necessary in any case.)

pjsier:
@james-minton sounds good! Also, it's not clear from the gif, but the one thing I did change in terms of a disabled style is remove the hover outline color change for the button. Otherwise it's the same though with the popover added

Lane:
@pjsier is this one good to merge or are there still revisions?

pjsier:
@lane this is good to go. We might update the disabled style later but this is good for now

PR #711, opened 2018-02-26
v0.9.1 - End revisions milstone

Lane:


pjsier:
Should we also change the version number on the package-lock.json file? Not sure if it matters, but I vaguely remember running into something weird with that before

PR #734, opened 2018-02-28
Condense header on scroll

Lane:
Drops the header from 80px to 56px on scroll.  Also adjusts so `header.scss` has the base header site-wide styles and `header.component.scss` has the style overrides for the map tool.

Adds a `scrolledToTop$` observable that emits true when the user scrolls into the top 56px of the page, and emits `false` when they scroll past that.


Lane:
whoops, forgot gif:
image

PR #740, opened 2018-03-01
Fix card order

Lane:
I thought this would be a quick fix but it took me a looonnng time to track down! The `activeFeatures` were being passed to the data panel, which sorts the location with `.sort()` and causes the original array to be modified in place.  Switched so that we sort a copy of the locations instead of the original array.


pjsier:
Ahhhhh this is totally my bad. I didn't think about the fact that `sort` modifies in place, and I thought it was only a one-way binding

Lane:
No worries! I didn't know it mutated the original until I looked it up, javascript kinda sucks for that sometimes.

PR #745, opened 2018-03-03
Top Evictors

Lane:
- Breaks the evictions an evictors list into child components of `ranking-tool.component`.
  - Sets up routes to allow navigation to top evictors

Still need to re-implement scroll to top, and put in the UI panel for top evictors.

### DIFF
--- a/src/app/app.component.ts
+++ b/src/app/app.component.ts
@@ -156,13 +156,13 @@ export class AppComponent implements OnInit {
   onLanguageSelect(lang) {
     this.translate.use(lang.id);
     this.analytics.trackEvent('languageSelection', { language: lang.id });
-    this.router.navigate([], {
-      relativeTo: this.activatedRoute,
-      queryParams: {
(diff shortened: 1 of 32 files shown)


Lane:
Alright this should be ready to go, changes since the PR was created include:

- Refactor social buttons into a shared component (`ui-social-share.component.ts`)
- Change `ranking-list.component` so it can show a list with different object properties
- Add `ranking-ui` to the top evictors list (search still to be implemented once data comes in)
- Add mock data for top evictors
- Add functions to `ranking.service` to handle sorting / filtering of top evictor data

pjsier:
I'm not sure we can address this, but it looks like it's reloading the data for cities or evictors from scratch each time I toggle between them. Can we set it up so that it only loads it on the initial load of the overall rankings component?

Lane:
yep, good catch.  That's an easy fix and should be in the branch now.

Issue #644, opened 2018-02-16
Separate meta data for ranking / map tool


Lane:
Have separate meta data for rankings and map tool so we can customize social share messages.

Issue #332, opened 2017-12-20
Setup Travis CI for Eviction Lab website


Lane:
Add Travis CI build to eviction-lab-website repository to build and push to S3 on all branches, and to invalidate the website cloudfront distribution on pushed to master.

pjsier:
Will be closed by https://github.com/EvictionLab/eviction-lab-website/pull/76

Lane:
Dropping this, can be tracked in the website repo now that we're using that

Issue #194, opened 2017-11-27
Support prototypes with non-hash URL?


pjsier:
Looking in the Angular routing docs for a different issue, they have [a code snippet](https://angular.io/guide/router#set-the-base-href) that supports using HTML5 routing on a dynamic base href. Could be worth looking into, and potentially triggering with an environment variable for prototype builds

Lane:
Sorry, missed this until now.  Is there any benefit of using non-hash routing over the hash routing?

pjsier:
Not sure, but I remembered that it's the way we originally started out, so I figured I'd bring it up. If you don't think it's a priority we can close it

Lane:
I'm fine with either way! Without the hash definitely looks nicer, and could potentially fix the mobile firefox jumping issue if it's caused by hash updates.  I think if we have time it's worth trying out.

pjsier:
I looked a little into this, and the big problem seems to be loading the assets with the `base` tag. Right now we're setting it with `href="./"`, which works because we're using the hash route. When I switched it to non-hash routing, it fails if there are any URL params present.

Since we're planning on the production deployment being at `evictionlab.org/rankings` and `evictionlab.org/map`, we'll probably need the base route to stay relative so we'd keep running into this issue. I'm sure there's a way of doing it, but I'm more skeptical that it's worthwhile now

pjsier:
Was thinking through this a bit, and I think we can do this in a pretty straightforward way, partially relating to #644. If we set S3's error template to index.html as well as the index template, then it'll forward all routes that fail to index.html without needing the hash route.

For Cloudfront we can have two origins for the tool (like we'd been talking about already), one that responds to `/map` and one that responds to `/rankings`, as well as S3 buckets for each of them. We can deploy the main site to the `/map` bucket and set its base href to `/map` with associated map metadata. Then, we can modify the metadata in the index.html file for rankings, but leave the base href the same, and deploy that file alone to the rankings bucket. It should load the assets from the map bucket so that we don't need to duplicate those, but we can still have the separate metadata.

Not going to work on this until we start putting the production deployment together, but hopefully it's a way of knocking out both issues

pjsier:
Closing this for general URLs with #753. Removed prototypes from Travis since we aren't really using it and it was pretty slow anyway

PR #753, opened 2018-03-06
Set up production deployment at sub paths

pjsier:
Adds script and functionality for deploying the tool to /map and /rankings with separate metadata (closing #644) while sharing a common set of cached assets. Also allows for using non-hash routing (closing #194) by redirecting all errored requests to S3 buckets to the bucket's index document. Also relates to #332 since the map tool is part of the website Cloudfront distribution. I'll need to update the Cloudfront ID env vars in Travis once we're ready to go ahead with this

This may take some tweaking, and we might have to ask someone about the SEO implications of returning 404s for tool sub-paths, but it should be fine for /map and /rankings which are the only pages with unique metadata anyway

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,38 +25,15 @@ before_deploy:
 - export PATH=$PATH:$HOME/.local/bin
 deploy:
 - skip_cleanup: true
-  provider: s3
-  access_key_id: $AWS_KEY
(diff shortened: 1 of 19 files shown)


Lane:
This is great! We might adjust the `/map` and `/rankings` URL paths later, but this works for now.

PR #755, opened 2018-03-06
Fix footer overlapping dropdowns

### LINKED ISSUE #729: States list causes page bounce on click, when page scrolled down (rankings)
states list bounces

This only happens when the page is scrolled down a bit (to where the footer is revealed) in states with a really short rankings list, e.g. large cities in West Virginia. Seen on Mac / Chrome / Desktop.

### LINKED ISSUE #730: States dropdown disappears into footer (rankings)
states list disappears into footer

Similar to #729 , this only occurs with shorter lists.

Lane:
Sets a max height on the ranking ui panel dropdowns so they don't go into the footer.  Fixes #730 and also seems to fix #729 

image

### DIFF
--- a/src/app/app.component.scss
+++ b/src/app/app.component.scss
@@ -12,5 +12,5 @@ app-header-bar {
 app-footer {
   display:block;
   position:relative;
-  z-index:997;
+  z-index:990;
 }
--- a/src/app/ranking/ranking-tool/ranking-tool.component.scss
+++ b/src/app/ranking/ranking-tool/ranking-tool.component.scss
@@ -93,7 +93,7 @@ $panelHeightLg: 194px;
 @media(min-width: $gtMobile) {
   .nav-bar {
     position: sticky;
-    z-index:30;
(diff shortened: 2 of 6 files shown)


pjsier:
It looks like this is still an issue for the Rank By dropdown

screen shot 2018-03-06 at 12 20 25 pm

Lane:
Ah, yeah I should've checked that one! Okay, new fix is in that adjusts the panel z-index when the dropdown is open.

pjsier:
The fix works great! I'll merge once the tests pass.

Also, I don't think this is related, but is scroll weirdly slow for you on this page? Not sure if there's an event listener that needs to be throttled somewhere, but the page is a bit jumpy to me

Lane:
Yeah, I've noticed it's slow when running on localhost, but it seems fine when running on a production URL.  Is it slow for you in production?

pjsier:
Nope, seems fine! In that case not an issue, just a little odd

PR #781, opened 2018-03-09
Add simple app loading shell

### LINKED ISSUE #712: App shell on initial load
Since we're loading a good amount of resources, should we set up an app shell to indicate to users that the page is loading? https://blog.angular-university.io/angular-app-shell/

I imagine we could just use the header and something like the existing loading spinner without changing much of the design, but that's an open question

pjsier:
Closes #712. I looked more into the Angular universal app shell, and it still looks like it would be rendering more than we need. It also seems like it might require building on a per-route basis. I'm still open to digging into it more, but I put together a sample just editing the `index.html` file with `app-progress-bar` code (since it's unlikely to change) and a simplified `app-header-bar` and it seems to work out fine. Let me know what you think!

Here's how it looks at a mobile resolution loading on Chrome with the Fast 3G preset and caching disabled. Desktop looks like same, but wanted to pull this as an example since the header changes more
app-shell

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -45,6 +45,60 @@
   <meta name="twitter:title" content="Eviction Lab: Map and Data" />
   <meta name="twitter:description" content="We’ve built the first nationwide database of evictions. Use our customizable map to find and compare eviction rates in your neighborhood, city, or state." />
   <meta name="twitter:image" content="https://beta.evictionlab.org/tool/assets/images/el-twitter-img.jpg" />
+
+  <style>
+    app-progress-bar {
+      position: fixed;
+      top: 0;
+      z-index: 1000;
+      position: absolute;
+      left: 0;
+      right: 0;
+      height: 5px;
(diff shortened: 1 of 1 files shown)


Lane:
Trying to test this out locally... the dev build gives me a really huge eviction lab logo before the app loads.  I can't run `--prod` builds locally since the switch to `beta.evictionlab.org` because I keep getting CORS errors.

image

pjsier:
Hmm I haven't run into this, but I'm not sure I tried it on different builds so I'll see if I can figure it out

Lane:
`ng build --env=dev --prod` works, thanks for that.  Not seeing the weirdness with the logo anymore, so it must have been related to that.

It all looks good to me!  Are there still some things to do on this or is it good to merge in?

pjsier:
If keeping it minimal like this is good with you then I think it's all good to merge!

PR #782, opened 2018-03-09
Fix map tool keyboard nav

### LINKED ISSUE #743: Update tab navigation, accessibility
Not sure if this happened during one of the refactoring PRs, but it doesn't look like tab navigation is still working on the map tool

Lane:
Updates focus styles and `tabindex` attributes so keyboard navigation flows in a logical order and every tabbed item has a focus state.

Closes #743

### DIFF
--- a/src/app/map-tool/location-cards/location-cards.component.html
+++ b/src/app/map-tool/location-cards/location-cards.component.html
@@ -21,7 +21,7 @@
         </span>
         <span *ngIf="prop.id !== 'divider'" class="card-stat-label">
           {{ prop.name }}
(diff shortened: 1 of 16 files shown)


pjsier:
This is working really well! Not sure if you want to split this into a separate issue, but the `mailto` link is skipped in the data panel. Otherwise everything on the site seems to work out

mailto-skipped

Lane:
Ah, firefox thing.  One sec, should be a quick fix.

pjsier:
Confirmed that this is a Firefox on Mac issue controlled by user settings, so this is good to go once the tests pass https://stackoverflow.com/questions/11704828/how-to-allow-keyboard-focus-of-links-in-firefox

Issue #620, opened 2018-02-15
Rankings accessibility implementation


pjsier:
Adding as a placeholder since this was mentioned but it's not currently in the issues

PR #784, opened 2018-03-09
Add keyboard nav to rankings

Lane:
Progress on #620 

- Makes sure all elements on page can be accessed via keyboard
- Fix positioning on scroll to top button
- Fix panel background not extending to the edge of screen on some resolutions.


Lane:
Added the following:
- Opening the panel moves keyboard focus to the close button in the panel
- Closing the panel restores focus to the item in the list that was selected
- Scrolls to list items when they are selected
- Esc closes the panel
- Left / right arrows go to previous / next items
- Bit of refactoring to clean things up

Should be good to merge.

PR #785, opened 2018-03-10
Refactor tile data fetching

pjsier:
I've been looking into the fact that eviction data is duplicated across the shape features and the centers (because initially it was difficult to combine them with the way the app is set up), and I've got it working now in a way that lets us remove all eviction data properties from the shape layer and reduce the tile size by about 15% (in local tests)

* Determines the `queryZoom` for the Mapbox tiles based on layer and area. This is necessary because the most detailed layer at zoom 10 won't always have the centers layer in the tile (even though hypothetically it should). By decreasing the zoom a bit for layers with larger shapes, we make sure that the tile will always have the centers layer.
* Uses GEOID instead of layer in URL query params for location, pulling layer from the GEOID length. It can get relatively long for tracts and block groups, but "block-groups" is the same length as block group GEOID, so it shouldn't be too bad. This is necessary because decreasing the zoom increases the amount of features returned, and it's a much faster and simpler way of making sure we get the right feature than some of the geography queries we were doing
* Queries the map for the centers data and merges it with the feature data on click to preserve the same experience of an immediate update

I've tested this locally by adding some lines that remove all eviction properties on the shape feature so that it's dependent on the centers layer, and everything seems to be working out. This should hopefully be a noticeable performance boost at least for load time

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -472,19 +472,6 @@
         }
       }
     },
-    "@turf/inside": {
-      "version": "4.7.3",
-      "resolved": "https://registry.npmjs.org/@turf/inside/-/inside-4.7.3.tgz",
-      "integrity": "sha1-5KhJafKIaJGzQ7Ebg/2jCV+6VCw=",
-      "requires": {
-        "@turf/invariant": "4.7.3"
-      }
-    },
-    "@turf/invariant": {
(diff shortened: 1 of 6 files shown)


Lane:
awesome, good thinking!

PR #800, opened 2018-03-13
move site nav to environment

Lane:
Moves the site navigation from `app.component` into the environment files.  Updates the language for "The Problem" and "Our Methodology".


pjsier:
This works, but if it's alright with you can we change the link to `/eviction-rankings` to `/rankings`?

Lane:
yep, makes sense.  Just updated.

PR #802, opened 2018-03-13
Feature/unsupported

### LINKED ISSUE #136: Notification for users with older browsers
Need to create and style a component for users with browsers that don't support WebGL with a link to information on downloading newer browsers

Lane:
Closes #136 
- Adds a notification on IE9 and below (as per HTML5 Boilerplate)
- Adds a dialog that pops up on the map tool if WebGL is not suppported.

IE9 and lower notification:
image

No WebGL dialog:
image


src/app/map-tool/map-tool.component.ts
```diff
@@ -121,7 +121,7 @@ export class MapToolComponent implements OnInit, OnDestroy, AfterViewInit {
 
   /** Checks if the map features are supported (currently just WebGL) and shows a dialog if not */
   checkSupport() {
-    if (!this.platform.hasWebGLSupport) {
+    if (this.platform.hasWebGLSupport) {
```
pjsier: Ah it looks like this slipped in there

Lane: ahhh yeah definitely don't want that

PR #817, opened 2018-03-15
WIP: Attempt to make null bubbles more obvious w/border

pjsier:
So I looked into addressing #812, and a couple issues came up. Currently, we're distinguishing null data from very low numbers by not showing bubbles at all. The problem with this is that our borders are thin, so it's not very easy to see the difference:
screen shot 2018-03-15 at 4 43 10 pm

Increasing the border size makes this difference significantly easier to see, and overall I think it addresses the problem:
screen shot 2018-03-15 at 4 34 55 pm

The problem, is using a white border to show null data can actually make circles more difficult to see when there isn't a choropleth layer selected, or even when the choropleth shading is significantly lighter. It ends up hiding some of the data in a way:
screen shot 2018-03-15 at 4 34 42 pm
screen shot 2018-03-15 at 4 45 34 pm

I know we've had the white borders on the bubbles for a while, but I think it's worth considering making them a darker color, or even the same color as the bubbles so that we can use them to distinguish from null data while not hiding lower results. Definitely want feedback from James and Noele (when they're both back) on this, but I think we could make things more clear.

Here are some examples. I'm not attached to the specific colors, mostly just showing them for contrast.

Borders the same color as the circles:
screen shot 2018-03-15 at 4 50 32 pm
screen shot 2018-03-15 at 4 50 25 pm

Borders slightly darker than the circles:
screen shot 2018-03-15 at 4 39 44 pm
screen shot 2018-03-15 at 4 39 51 pm


pjsier:
Closing in favor of #823

PR #819, opened 2018-03-16
WIP: Add null data pattern to legend

### LINKED ISSUE #788: Add null data pattern to legend
We don't currently have a place indicating that the striped pattern means no data

pjsier:
Closes #788 pending design feedback. Uses CSS `repeating-linear-gradient` instead of the base SVG pattern so that we don't have to worry about the iOS and Safari compatibility issues

screen shot 2018-03-16 at 8 21 35 am

screen shot 2018-03-16 at 8 21 49 am


noelelusano:
Design forthcoming.
Let's use "Unavailable" in place of "N/A" to match what it is in the Map cards

pjsier:
Closing in favor of #861

PR #829, opened 2018-03-19
Only fire moveend when scroll enabled, default fills

### LINKED ISSUE #810: Scroll is jumpy on iOS Chrome, Firefox
Thought this was fixed by #803, but still seeing it a bit. The fixed mobile header gets pulled under the URL bar, and then reappears, causing a scroll jump. Not sure what the original cause is

pjsier:
Closes #810. We're delaying the height transition of the map using viewport units on mobile because of the URL bar appearing and disappearing, but the window is still changing size whenever that happens. That changes the dimensions of the map which then triggers a `moveend` event to update the `bounds` URL parameters. This may have started behaving differently when we updated routing, but updating those parameters makes the URL bar reappear and causes a jumpy scroll effect

This PR only fires the bounds update event when `scrollZoom` is enabled, so that it's not happening when someone is scrolling down the page. I'm also adding defaults for `fill-color` on the map layers, because I noticed an occasional black flicker before it's updated the first time

### DIFF
--- a/src/app/map-tool/map/map/map.component.ts
+++ b/src/app/map-tool/map/map/map.component.ts
@@ -359,9 +359,11 @@ export class MapComponent implements OnInit, OnChanges {
     this._store.bounds = this.mapService.getBoundsArray()
       .reduce((a, b) => a.concat(b))
       .map(v => Math.round(v * 1000) / 1000);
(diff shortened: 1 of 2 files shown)


Lane:
Ah good catch.  MY mobile browsers don't do the shrink on Android so I wasn't seeing this.

PR #861, opened 2018-03-22
Add bubbles to legend

Lane:
Adds bubbles and no data to legend.

@pjsier not quite sure how to pull to top level of the scale from the bubble attribute for the labels.  Could you help me with that?

image


pjsier:
Yep, will do!

pjsier:
Just updated with the interpolation. Also, it looks like there are some bugs with positioning on Firefox. It usually looks right at first, but then after it changes a few times I get this:

screen shot 2018-03-21 at 8 30 55 pm

Also, question for @noelelusano, should we use the same color we're using for the bubbles on the map?

Lane:
Hmm.. I can't replicate it jumping out of alignment like you have in your screenshot.  Is there a specific action that seems to cause it to happen?

pjsier:
That's weird, it happens to me usually when first selecting a bubble property, but changing the bubble property fixes it. Updating the zoom breaks it again though

Here's an example:
bubbles

Lane:
Definitely weird, not happening on latest OSX Firefox for me.  Looking at the CSS though, I _think_ I might know what's causing it.  I just pushed a potential fix, could you try it out to see if it fixes it?

pjsier:
Yep, that fixes it! Weird that it didn't show up for you though

PR #871, opened 2018-03-22
Fix IE11 Map Not Loading

Lane:
Closes #834 where the map wouldn't load on every second page view.  Initializing the map with an empty style, then setting the style immediately after seems to fix it.

PR #876, opened 2018-03-23
Loading Fixes

### LINKED ISSUE #843: IE11: Loading bar showing indefinitely (poor performance)
- **Device/Browser**: Windows 10/IE 11
  - **Resolution:** 1280x720
  - **Problem**: The map loads, but the loading bar repeats on an infinite cycle of showing for a few seconds, then disappearing for a few hundred milliseconds.  Scrolling performance is also noticably bad while this is happening.  Browse window will become unresponsive if left long enough.  Potential repeating task eating up CPU cycles? 
  - **Expected**: The map should load, the loading bar should disappear, there shouldn't be any map related tasks running in the background.
  - **To Reproduce:** 
    - Load the following URL :
  https://beta.evictionlab.org/map/2014?geography=counties&bounds=-120.912,31.942,-80.246,46.912&choropleth=rb&locations=08077,-108.219,38.933%2B20193,-101.056,39.351%2B08099,-102.395,37.956
    - Wait for everything to finish rendering
    - Loading bar will begin its infinite cycle
    - Leave it for a minute or two, map shapes will disappear except for selected location outlines

Lane:
- Changes the loading service implementation to be observable based with timeouts.

You can load items with `start()` and `end()` as done previously, or pass an observable that resolves when loading is complete.  All loading items will timeout after 10 seconds.  

- Adds debugging that logs how long each loading task takes. (can be shut off with `_debug` attribute)

image

- Fixes loading loop on browsers that are slow to render (closes #843)

### DIFF
--- a/package.json
+++ b/package.json
@@ -6,8 +6,8 @@
     "ng": "ng",
     "start": "ng serve",
     "build": "node ./build/pre-build.js && ng build",
-    "build-prod": "node ./build/pre-build.js && ng build --prod --env=prod --base-href=/map/ --deploy-url=https://beta.evictionlab.org/tool/",
-    "build-dev": "node ./build/pre-build.js && ng build --prod --env=staging --base-href=/map/ --deploy-url=https://staging.evictionlab.org/tool/",
+    "build-prod": "node ./build/pre-build.js && ng build --prod --env=prod --deploy-url=https://beta.evictionlab.org/tool/",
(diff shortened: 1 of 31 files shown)


Lane:
Oh crap, I branched this off of #871 😬

pjsier:
No worries! I think we can just drop the other one and merge this in

Lane:
works for me!

PR #879, opened 2018-03-24
Change census switch method

### LINKED ISSUE #853: IE11: Crash on moving year slider
- **Device/Browser**: Windows 10/IE 11
  - **Resolution:** 1423x800 (3200x1800 @ 225% zoom)
  - **Problem**: When moving the year slider to a year that has to load new tiles, the browser window becomes unresponsive, and prompts to "stop long running script".  Stopping the script turns all bubbles black.  The new data will eventually load, but the browser can be unresponsive for up to a minute.
  - **Expected**: The bubbles should update without locking up the browser.
  - **To Reproduce:** Load a view with a lot of bubbles, move the year slider from a post 2010 year to a pre 2010 year or vice versa.

image

Not sure there is much we can do to improve performance when switching to a year that needs a new tileset, but maybe we could show a warning if there is no way to improve performance.

Lane:
Previously we were using `map.setStyle` to set a new style when the census year changed, which IE or mobile browsers didn't handle too well.  This PR changes that so all of the layers using old source data are removed, and then re-added with the new source.  The choropleths / bubbles are removed momentarily, but the browser doesn't lock up at all and the update is pretty quick.

Fixes #853, and I think this is the end of our IE performance issues! 🎉 🤞

### DIFF
--- a/src/app/map-tool/map/map.service.ts
+++ b/src/app/map-tool/map/map.service.ts
@@ -299,15 +299,23 @@ export class MapService {
   updateCensusSource(layerGroups: MapLayerGroup[], sourceSuffix: string) {
     const mapStyle: mapboxgl.Style = this.map.getStyle();
     const layerPrefixes = layerGroups.map(l => l.id);
-
-    mapStyle.layers.map(l => {
-      const layerPrefix = l.id.split('_')[0];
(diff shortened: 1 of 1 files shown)


pjsier:
This is great!

PR #880, opened 2018-03-24
Feature overview fallbacks for IE / Edge

### LINKED ISSUE #836: IE11/Edge: Feature Overview Problems
- **Device/Browser**: Windows 10/IE11/Edge
  - **Resolution:** 1280x720
  - **Problem**: A couple:
    - IE11/Edge: Animations do not display properly (missing vendor prefixes?)
    - IE11: Text doesn't wrap and instead overflows into next slide

Lane:
IE / Edge do not support "transform" styles via CSS on SVG elements.  This adds a check for SVG CSS transform support and uses animated gifs if there is no support.

Closes #836


pjsier:
One thing worth checking, I've run into some cases where IE 11 doesn't auto-play gifs, is this running on BrowserStack? Either way, the first frame of the gif is better than nothing

Lane:
Didn't check on BrowserStack, but checked on IE11 on my Windows 10 laptop and they were auto playing.

pjsier:
Good enough for me!

PR #893, opened 2018-03-26
Re-order `<link>` and `<meta>` tags

### LINKED ISSUE #635: Multiple favicon requests
Getting a lot of HTTP requests for the same favicon, should try and limit it to one request so that these requests don't block other requests.

image

Lane:
Adjusts the ordering of meta and link tags, so that meta tags come first to match structure of [HTML5 boilerplate](https://github.com/h5bp/html5-boilerplate/blob/master/src/index.html).  Also removes favicon.ico `<link>` tag, as browsers will automatically pull that from the root directory.

This clears up the multiple requests for the favicon pngs on resize and closes #635

### DIFF
--- a/src/index.html
+++ b/src/index.html
@@ -2,33 +2,12 @@
 <html lang="en">
 <head>
   <meta charset="utf-8">
+  <meta http-equiv="x-ua-compatible" content="ie=edge">
   <title>Eviction Lab: Map and Data</title>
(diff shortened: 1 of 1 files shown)


pjsier:
Will the root directory be "https://evictionlab.org/favicon.ico", meaning we only have to include the favicon info on the website repo?

PR #907, opened 2018-03-27
Make mobile home link functional

pjsier:


Lane:
Wow, thank you, this would have totally slipped my mind.

PR #932, opened 2018-03-28
Fix Ranking Overflow

Lane:
Closes #932   
  - Adjustments so that the ranking panel does not need a fixed height 
  - Limiting long titles to one line
  - Moves scroll to top button inside of `evictions.component`
  - Set offset to show scroll to top button to a fixed number (we were detecting dynamically, but it's not really needed and this is better for performance)

Mobile: 
image

Tablet portrait:
image

### DIFF
--- a/src/app/ranking/ranking-panel/ranking-panel.component.scss
+++ b/src/app/ranking/ranking-panel/ranking-panel.component.scss
@@ -295,6 +295,7 @@
 }
 
 .close-button {
+  border-radius:0;
   cursor: pointer;
   text-align: center;
   padding: grid(1.25);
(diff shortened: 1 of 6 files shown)


src/app/ranking/ranking-panel/ranking-panel.component.scss
```diff
@@ -139,12 +144,18 @@
...
   .rank-location {
     @include altSmallCapsText(26px);
+    @include ellipsis-overflow();
     text-align: left;
     span {
       @include smallCapsText(14px);
     }
   }
 }
+@media(min-width: $gtTablet) {
```
pjsier: I'm guessing this is left over from testing?

Lane: yep, very non-essential, will fix

PR #940, opened 2018-03-29
Location Card Updates

### LINKED ISSUE #918: Drop transitions on data panel location cards
It would be better for performance if we dropped the transitions from the data panel location cards.  Currently they are transitioning at the same time as the map cards which can sometimes cause a delay.  The transitions aren't really appropriate for the data panel either.

### LINKED ISSUE #936: Long city names overflow or cutoff in location cards
I went through the list of longest names in the census data, and found a decent amount of cities that overflow the map card and get cut off in the location cards. Not sure if there's a way of setting something like `overflow: ellipsis` after a certain number of lines, or if we should allow them to expand their containers

In the map:
screen shot 2018-03-29 at 2 49 54 pm

In the data panel (should be "Manderson-White Horse Creek"):
screen shot 2018-03-29 at 3 07 25 pm

Here's a list of cities I found (most are accessible via search):

* University of Pittsburgh Johnstown
* Fetters Hot Springs-Agua Caliente
* Bonadelle Ranchos-Madera Ranchos
* University of California Merced
* Village of Grosse Point Shores
* Lexington-Fayette Urban County
* Conning Towers Nautilus Park

Only cut off in data panel:

* Encantada Ranchito-El-Calaboz
* Shorewood-Tower Hills-Harbert
* Pocono Mountain Lake Estates
* Cusetta-Chattahoochee County
* Manderson-White Horse Creek

Lane:
- Fix location name overflow by truncating and adding tooltip for long names (closes #936 )
  - Drop transitions for data panel cards (closes #918 )

image

### DIFF
--- a/src/app/map-tool/location-cards/location-cards.component.html
+++ b/src/app/map-tool/location-cards/location-cards.component.html
@@ -5,7 +5,10 @@
 >
   <div class="card-header" [class.clickable]="clickedHeader.observers.length > 0" (click)="clickedHeader.emit(f)">
     <app-ui-icon class="marker" icon="marker"></app-ui-icon>
-    <h1 class="card-heading">{{f.properties['n']}}</h1>
+    <h1 class="card-heading" 
+      [tooltip]="f.properties['n'].length > maxLocationLength ? f.properties['n'] : null" 
+      [placement]="collapsible ? 'top' : 'bottom'"
+    >{{ getLocationName(f.properties['n']) }}</h1>
     <h2 class="card-subheading">{{f.properties['pl']}}</h2>
     <app-ui-close-button class="card-dismiss" [ariaLabel]="'DATA.REMOVE_LOCATION' | translate:{'name':f.properties.n}" (onPress)="dismissedCard.emit(f)"></app-ui-close-button>
   </div>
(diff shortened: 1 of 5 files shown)


pjsier:
It looks like some are still getting cut off in the data panel (should start with "Bonadelle")

screen shot 2018-03-29 at 5 08 55 pm

pjsier:
Also, it looks like this is dropping transitions for all location cards, not just on the data panel?

Lane:
Oh weird, will have to check transitions on Firefox.

I don't know if we can fix the cutoff in the data panel without some pretty big changes.  The space in the data cards only allows for 2 lines of text where the map cards allow 3.  Our options are: 

  - boost the header height to allow 3 lines of text, which brings in a lot of white space on smaller names.
  - drop the maximum number of characters for data panel cards
  - reduce font size for the header to allow 3 lines

pjsier:
The third option (reducing the font size) sounds like the best to me, but this sounds like a question for @noelelusano

Lane:
Okay, boosted height by 20px and it allows for 3 lines without too much white space.

@pjsier the animations seem to be working for me in the map cards on Chrome / Firefox / Safari.  Are you getting any errors logged or anything?

pjsier:
Great! And sorry, meant to clarify earlier, but sometimes they're just a bit delayed for me on Firefox. They work though

Issue #831, opened 2018-03-19
Safari: Map zoom not restored on scroll


pjsier:
- **Device/Browser**: macOS High Sierra/Safari 11
  - **Resolution:** 2560x1600
  - **Problem**: After viewing the data panel on Safari and then scrolling back up to the map, you're not able to zoom into the map with scroll unless you zoom out inside the map first:
![safari-bug](https://user-images.githubusercontent.com/8291663/37627279-bd0924f4-2ba1-11e8-89e5-f9efd5a46197.gif)

  - **Expected**:  You should be able to scroll to zoom both in and out of the map after scrolling to the top or clicking "Back to Map"
  - **To Reproduce:**  Select a location, then click "View more Data" or scroll down. Then scroll back up or click back to map, and try to zoom into the map by scrolling

I'm guessing this could have something to do with restoring context? But the fact that it works when zooming out it weird

pjsier:
It looks like this might actually be a general Safari issue. If I scroll down the page on this link https://www.mapbox.com/mapbox-gl-js/example/cluster/ and then scroll back up, the same thing happens

pjsier:
This doesn't show up in earlier versions of Safari (tested on 9) that don't have the elastic scroll, so I'm downgrading this to low priority since a browser issue for maps on Safari 11

PR #942, opened 2018-03-29
Prevent safari overscroll

Lane:
Adds a mousewheel listener in the scroll service that prevents the browser default when the user tries to scroll up, but their already at the top of the page.

This is as good of a fix as I can get for #831.  The map still won't zoom, the first time they try to zoom in but it works on the second try so it's better than nothing.  Tried to debug and find a way to get it to zoom on the first try, but it looks like the browser is just eating those events and not passing them on to the DOM.

### DIFF
--- a/src/app/services/scroll.service.ts
+++ b/src/app/services/scroll.service.ts
@@ -38,11 +38,21 @@ export class ScrollService {
     @Inject(DOCUMENT) private document: any,
     private platform: PlatformService
   ) {
-    // provide observable with top / bottom vertical offset
+    // kill overscroll, primarily for safari so the map is zoomable with
+    // mousewheel / trackpad.
+    this.platform.nativeWindow.addEventListener('mousewheel', (e) => {
+      if (e && e.deltaY < 0) {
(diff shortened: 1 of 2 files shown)


src/app/services/scroll.service.ts
```diff
@@ -38,11 +38,21 @@ export class ScrollService {
     @Inject(DOCUMENT) private document: any,
     private platform: PlatformService
   ) {
-    // provide observable with top / bottom vertical offset
+    // kill overscroll, primarily for safari so the map is zoomable with
+    // mousewheel / trackpad.
+    this.platform.nativeWindow.addEventListener('mousewheel', (e) => {
```
pjsier: Is this firing on every single event? Or only the first time?

Lane: It's firing for every single one, because `preventDefault` needs to be called on all of them to prevent the overscroll from happening.  There's not much overhead going on in the handler so it shouldn't really effect performance, but I could do a user agent check and only set it for Safari.

PR #956, opened 2018-03-30
Cap Y axis at 100%, display >100% above

### LINKED ISSUE #944: Cap graph Y axis at 100%
We're displaying any data over 100% in the location cards as ">100%", but we need to find out if it's alright to display it as is in the graph, which we're currently doing

pjsier:
Closes #944, dependent on https://github.com/EvictionLab/angular-d3-graph/pull/16. Caps the Y axis at 100%, displays any values above at the top of the axis with ">100%"

cap-axis

Fixed on the bar chart (set the max value to 50% to make it more obvious)

screen shot 2018-03-30 at 9 06 11 pm

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -931,9 +931,9 @@
       "integrity": "sha1-SlKCrBZHKek2Gbz9OtFR+BfOkfU="
     },
     "angular-d3-graph": {
-      "version": "2.0.13",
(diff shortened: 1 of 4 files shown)


pjsier:
It looks like `yPos` isn't working with the bar. I'll take a closer look

Issue #901, opened 2018-03-26
Get More Data form updates


pjsier:
Adding as a placeholder

PR #963, opened 2018-03-31
Changes route to data download to Lambda

pjsier:
Progress on #901. I deployed a Lambda that takes the `email` parameter from a submitted form, POSTs it to MailChimp, and then returns a 302 redirect (currently to preview.evictionlab.org). I haven't changed any of the other MailChimp pieces so far


Lane:
Good to close this?  Looks like the dialog is not needed anymore, as the form will be on the "Get the data" page.

pjsier:
Works for me! Closing

PR #968, opened 2018-04-02
Fix layer ordering

### LINKED ISSUE #967: Census switch moves bubble layers in front of labels
It looks like a side effect of #878 was that bubble layers are moved on top of labels:

screen shot 2018-04-02 at 7 50 20 am

pjsier:
Closes #967. The currently `style.json` doesn't really require any logic to determine where to add layers, so I moved `counties_text` to the top since it rarely appears and should take precedence anyway, and set a static `addBeforeLayer` to make sure all bubble layers are added before the city labels

screen shot 2018-04-02 at 7 55 14 am

### DIFF
--- a/src/app/map-tool/map/map.service.ts
+++ b/src/app/map-tool/map/map.service.ts
@@ -33,6 +33,7 @@ export class MapService {
   private _mapHighlights = [];
   private _mapHover = [];
   private _hoverEnabled = true;
+  private addBeforeLayer = 'city_extra_small_labels';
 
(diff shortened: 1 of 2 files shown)


Lane:
I think we still need a function to determine where the layers get inserted.  The choropleth layers get inserted above the roads and outlines after a census switch:

image

pjsier:
Ah good point, I'll update this

pjsier:
The latest commit should fix it!

PR #972, opened 2018-04-02
Update footer styles

Lane:
Updates footer to match new mockups.

Mobile:
image

Tablet:
image

Desktop:
image


src/environments/environment.ts
```diff
@@ -34,5 +34,18 @@ export const environment = {
     { url: '/methods', langKey: 'NAV.METHODS' },
     { url: '/help-faq', langKey: 'NAV.HELP' },
     { url: '/updates', langKey: 'NAV.UPDATES' }
+  ],
+  footerNav: [
+    { url: '/', langKey: 'NAV.HOME' },
+    { url: '/map', langKey: 'NAV.MAP' },
```
pjsier: Small thing, but this would be `/` on dev, right?

src/assets/i18n/convert-to-csv.js
```diff
@@ -0,0 +1,218 @@
+const data = {
```
pjsier: Are we thinking of using this again? If we are I think it's fine to add

Lane: oh whoops, didn't mean to add this in

PR #979, opened 2018-04-02
Print styles

Lane:
Add some overrides so the page content is printable.  Not a pretty implementation, but seems to be okay.

Safari / Firefox do not show the map when you go to print, but Chrome does.  Read an issue that said Safari / Firefox will only print when `preserveDrawingBuffer` is true, but I think I read that could cause performance issues?

Here are some sample PDFs from printing:
[Eviction Rankings _ Eviction Lab.pdf](https://github.com/EvictionLab/eviction-maps/files/1869397/Eviction.Rankings._.Eviction.Lab.pdf)
[Eviction Map & Data _ Eviction Lab.pdf](https://github.com/EvictionLab/eviction-maps/files/1869398/Eviction.Map.Data._.Eviction.Lab.pdf)


pjsier:
Yeah adding the preserve drawing buffer setting definitely causes performance issues, so this seems good to me

Lane:
Sounds good, added a message that shows up if the map can't print.  This should be ready.

PR #1011, opened 2018-04-03
Add state fallback tweet, mention area type

### LINKED ISSUE #1008: Rankings should use general tweet if state data is unavailable
Alaska's tweet currently comes out similar to "-1.0% of renter homes were evicted in Alaska in 2016. Badger had the highest rate." We should use the general area type tweet if the state data is unavailable

### LINKED ISSUE #1009: Rankings state tweet should mention area type
An example tweet when a state is filtered for but no location selected is "1.82% of renter homes were evicted in Alabama in 2016. Gadsden had the highest rate." It doesn't mention that Gadsen has the highest rate out of mid-sized cities. We'll need to add this and update the translations doc

pjsier:
Closes #1008 by adding a fallback tweet that just includes the link

Closes #1009 by adding the area type to the region tweet


Lane:
Looks good, thanks for updating the content matrix!

PR #1026, opened 2018-04-04
set focus element after dismissing card

### LINKED ISSUE #1017: IE / Edge: Set focus element when dismissing location
When dismissing an element via keyboard in IE / Edge, focus is reset to the beginning of the document and requires you to tab through the whole page to get back to where you were.

Lane:
- Sets focus to another card's close button when dismissing a map card
- Sets focus to the first "Add location" input when dismissing a data panel card

Closes #1017


pjsier:
This works for me on Chrome and Safari, but not on Firefox

pjsier:
Looks like Firefox just needed a slightly longer delay, otherwise it would pick up the removed card. Just increased the timeout from 500 to 750 and that fixes it without breaking anything

PR #1032, opened 2018-04-04
Fix focus on dismiss

Lane:
Whoops, I missed that you had fixed the Firefox issue by bumping up the timeout delay.

I've adjusted the function so it doesn't rely on a timeout, because some devices could take longer to dismiss the card than others and it's not really reliable.  This also focuses the element in the proper order.


pjsier:
This works a lot better, and agreed that it's more reliable

PR #1046, opened 2018-04-05
Set tooltip position dynamically to keep visible

### LINKED ISSUE #1045: Tooltips at right edge of screen go out of view
We need to add a bit of logic to have the tooltip be on the left if it's the last property visible in the map cards or on the last card in the data panel

screen shot 2018-04-05 at 6 54 01 am

screen shot 2018-04-05 at 6 54 09 am

pjsier:
Closes #1045. Trying to be extra careful with this one. Sets the tooltip position to `top` instead of `right` if the screen size is smaller than tablet and either the card is the 3rd card, or the property is the last property within a card. There will still be situations when the tooltip will go out of view, but people will be able to scroll to bring it in view instead of it being totally inaccessible.

Also fixes an issue for the embed where longer legend tooltip descriptions could go out of view. This might not account for everything, but it shrinks it so most will fit.

screen shot 2018-04-05 at 7 27 57 am
screen shot 2018-04-05 at 7 28 05 am
screen shot 2018-04-05 at 7 27 39 am
screen shot 2018-04-05 at 7 33 19 am

screen shot 2018-04-05 at 9 14 45 am
screen shot 2018-04-05 at 9 14 26 am

### DIFF
--- a/src/app/map-tool/embed/embed.component.scss
+++ b/src/app/map-tool/embed/embed.component.scss
@@ -52,6 +52,14 @@ html.embedded {
   // fix disappearing legend
   .map-ui-wrapper { height: 100vh!important; }
 }
+
+// Shrink legend tooltip to make sure it fits
+.tooltip-inner {
+  padding: 4px;
+  font-size: 11px;
+  max-width: 360px;
+}
+
 @media(min-width: $gtMobile) {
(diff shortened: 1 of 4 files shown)


Lane:
Looks great, thanks for fixing this up.

PR #1064, opened 2018-04-07
Only show photo attribution on rankings

### LINKED ISSUE #1063: Update footer photo attributions

pjsier:
Closes #1063


Lane:
I think this is already done in #1020 ?

Lane:
Well, the showing only on rankings, not the credits update.  I guess this wouldn't hurt anything so I can merge anyway.

pjsier:
Let me double check, I thought it was showing up before for some reason

pjsier:
Yep, you're right. Not sure where I missed that. I updated this to only fix the attribution name

PR #1091, opened 2018-04-10
Add block group and census tract labels to cards

Lane:
- Moves the truncating for location name to map service.  Putting it here makes sure the truncation function is only called once instead of every change detection cycle.
  - Adds Block Group / Census Tract labels to the card when one is selected.

Seems to work okay for English, but with Spanish certain tracts / block groups will overflow the card.  Not sure the best way to handle this.

image

### DIFF
--- a/src/app/map-tool/location-cards/location-cards.component.html
+++ b/src/app/map-tool/location-cards/location-cards.component.html
@@ -9,10 +9,10 @@
       [tooltip]="isTruncated(f['displayName']) ? f.properties['n'] : null" 
       [placement]="collapsible ? 'top' : 'bottom'"
     >
-      <small *ngIf="f.properties['layerId'] === 'tracts' || f.properties['layerId'] === 'block-groups'">
+      <span *ngIf="f.properties['layerId'] === 'tracts' || f.properties['layerId'] === 'block-groups'">
         {{ f.properties['layerId'] === 'tracts' ? ('DATA.TRACT_SINGULAR' | translate) : ''}}
         {{ f.properties['layerId'] === 'block-groups' ? ('DATA.BLOCK_GROUP_SINGULAR' | translate) : ''}}
-      </small>
+      </span>
       {{f['displayName']}}
     </h1>
     <h2 class="card-subheading">{{f.properties['pl']}}</h2>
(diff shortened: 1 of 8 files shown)


pjsier:
Could we make the block group/tract label smaller and keep the actual ID the same size? Would need design input on that

Lane:
Yeah, I think that could work.  I'll make that change and then see if it's alright with Noele.

Lane:
Updates styles based on sketch file provided by Noele.

- Adjusts font sizes on map cards
- Switches to Akkurat for tract and block group label on tablet+ 
- Boosts min-width on cards in tablet portrait ( closes #1092 )

Prototype link: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/issue-1078

Lane:
This should be ready, here's a prototype link w/ block groups and tracts: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/issue-1078/#/2016?geography=counties&bounds=-87.987,34.524,-84.587,37.236&lang=es&locations=47185935400,-85.449,35.908%2B470370101052,-86.859,36.245%2B47125,-87.387,36.477

pjsier:
Testing on a couple browsers now, but one thing I wanted to note is that the tract/block group name shows up without the label in the the graph tooltips:

screen shot 2018-04-11 at 8 27 48 am

I actually think this is ok, and probably preferable, but I wanted to make sure everyone's on the same page about it. Related to this, I'd prefer to follow this sort of lead of initially including the label and then leaving it off later for the exports (https://github.com/EvictionLab/eviction-lab-exports/issues/71). Would it make sense to just include the full "Census Tract X" in the title of the PowerPoint and PDF, and then leave the name as is throughout each export?

Lane:
That sounds good to me.  @james-minton that alright with you?

pjsier:
Other than that everything seems to look good. Since I missed a lot of the style tweak conversations yesterday I'm just gonna list what I'm noticing as different to make sure it's right:

* On desktop, the tract/block group label should be on the line before the area name in a smaller font size with Akkurat
* On tablet, the tract/block group label should be the same font, font size, and line as the area name
* The stat labels in location cards on tablet and mobile should have `text-transform: uppercase` applied, but not on desktop

If that's all good then I think this is good to go

PR #1097, opened 2018-04-11
Add static geocoder for states and counties

### LINKED ISSUE #1081: Map geocoder fallback
Create static search with states and counties for map as a fallback in case of geocoder service disruptions.

pjsier:
Closes #1081. Implements a static file location search for states and counties with `StaticSource` when `useMapbox` is false. The build step for `locations.csv` is added to the ETL repo already. I also added the Mapbox text attribution back to the OpenMapTiles `style.json` because from what I remember we still needed that since we're using Mapbox GL JS, but we needed to remove the Mapbox logo.

### DIFF
--- a/src/app/services/search-sources.ts
+++ b/src/app/services/search-sources.ts
@@ -104,3 +104,27 @@ export const OSMNamesSource: SearchSource = {
         });
     }
 };
+
+export const StaticSource: SearchSource = {
+    key: '',
+    baseUrl: '',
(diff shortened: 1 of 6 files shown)


Lane:
Works great!  I'm not sure about the Mapbox attribution, on [their attribution page](https://www.mapbox.com/help/how-attribution-works/) it says:

> If your map does not use Mapbox designs, data, hosting, or other Mapbox APIs, Mapbox does not require you to provide attribution in either form.

PR #1105, opened 2018-04-11
Set height with javascript instead of using `vh`

Lane:
There are a few weird things going on with using `vh` units on mobile / tablet, due to the changing size of location bar on the browsers.

This sets the height explicitly by getting the height with javascript and setting it. Also removes the need for a bunch of device specific CSS.


Lane:
Closes #1105 

image

PR #1110, opened 2018-04-13
Fix setting mobile height

### LINKED ISSUE #1109: Skip first width change on mobile
I missed this in in #1105 and #1106, and thought it had to do with device specific padding, but we're firing `setMapSize` after one second, and then again on the initial width change even though `distinctUntilChanged` is set. We should skip the first change in the observable

pjsier:
Closes #1109. Right now, the mobile map height is always going to be a little off after the address bar shifts because that will fire `setMapSize`, and it won't get called again until the width changes. It might be good to get this in before the next release for that reason


Lane:
Good catch, just tested this out and is working as expected.  I added a version bump to `version.ts`  because I forgot I had to do that manually.  Merging into development, then will merge to master.

PR #1116, opened 2018-04-13
update spanish translations

Lane:
Adds in updates for
  - max locations error
  - census demographics label
  - ranking panel robotext

### DIFF
--- a/src/assets/i18n/es.json
+++ b/src/assets/i18n/es.json
@@ -63,7 +63,7 @@
     "SEND_DATA": "Enviar datos",
     "AUTOCOMPLETE_HINT": "Cuando esté disponible, use las flechas hacia arriba y hacia abajo para revisar e ingresar para seleccionar.",
     "LOADING": "Cargando",
-    "TRACT_SINGULAR": "Sección Censal",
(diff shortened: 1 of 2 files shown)


src/assets/i18n/es.json
```diff
@@ -63,7 +63,7 @@
     "SEND_DATA": "Enviar datos",
     "AUTOCOMPLETE_HINT": "Cuando esté disponible, use las flechas hacia arriba y hacia abajo para revisar e ingresar para seleccionar.",
     "LOADING": "Cargando",
-    "TRACT_SINGULAR": "Sección Censal",
+    "TRACT_SINGULAR": "Sección Censal",
```
pjsier: It looks like this is just a whitespace addition?

Lane: whoops, will drop

pjsier:
It looks like the Census Demographics translation goes right up against the borders of the table on mobile. Not sure how we want to handle that, but everything else looks good

screen shot 2018-04-13 at 1 59 12 pm

Lane:
image

Added some padding so it wraps to a second line.

PR #1155, opened 2018-04-25
Add link to FAQ in rankings intro

### LINKED ISSUE #1151: Add new language to Rankings Intro and link to FAQ item
Could we add the following language to the Rankings intro? Right at the end of the paragraph:

>(For more about the locations ranked here, see our FAQ.)

Translation:

>(Para obtener más información sobre los lugares clasificados aquí, consulte nuestras Preguntas frecuentes).

Here's the link to the FAQ item:

`/help-faq/#rankings-locations`

Lane:
Closes #1151 

image

image


james-minton:
Nice. While we're on the topic, have you ever thought it weird that it's "Cities in United States"? Most Americans would add the article "the" before U.S. in this case.

Lane:
I can add a check to see if the region is "United States" and put the word "the" in if that works.  My Spanish is bad, would we need to do something similar there?

james-minton:
That sounds good ... my Spanish isn't great either, but we can probably rely on our existing translation there? Thanks and I'm good with a merge whenever you are!

PR #1159, opened 2018-04-26
Add map link from rankings

### LINKED ISSUE #1147: Link to map from rankings
Add a link to the map in the rankings panel for a location.  Link should open in new tab.

pjsier:
Closes #1147. Adds a map link in the rankings panel that loads the map with the selected city, rankings year, and eviction rate as the bubble layer. We could also set the bounds to some arbitrary bounding box based on the location center, but not sure if it's necessary. If "view in map" is "ver en el mapa" in Spanish then this should be good to go. 

screen shot 2018-04-26 at 12 15 13 pm


Lane:
This looks good! I think setting the bounds to a box centered around the location is a good idea.  I will check with Abby on the translation.

pjsier:
Just updated it to set the view to an approximate bounding box (I'm using New York as the baseline for the center padding, so it should fit every city, even if it's more zoomed out from some)

PR #1161, opened 2018-04-28
WIP: Graph Embed

Lane:
An embeddable graph component that displays line graphs based on provided query parameters.  The query parameter `items` contains a list of locations and the corresponding property to graph.  For national numbers, use `national`, for all other locations use the format used in the map `{{geoid}},{{lon}},{{lat}}`.

**Example 1:**
National evictions vs. filings: http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/graph-embed/#/graph?items=nationwide|e+nationwide|ef

**Example 2:**
You can add a line for any location / data property.  For example, here is a not so useful graph of Richmond eviction rates vs. Clayton County poverty rates.
http://eviction-lab-prototypes.s3-website.us-east-2.amazonaws.com/graph-embed/#/graph?items=5167000,-77.476,37.53%7Cer%2B%2B13063,-84.358,33.542%7Cpr

Had to do a bit of refactoring to be able to re-use some components:
  - Graph tooltips is now its own component
  - Data retrieval is moved out of `map-tool.service` and is now in `data.service`

Looking for design feedback right now.  Not quite ready to merge, I still need to clean up a few things and update tests.


pjsier:
I think the layout makes sense, and I may have missed some things when I was out, but I'm wondering if we should add functionality like showing non-eviction properties (with the exception of the nationwide numbers). I'm not sure how the UI would work for creating a graph embed if there are additional features like this

Lane:
I don't think we have a plan to build a UI to create graph embeds right now. This will just be for in-house use on the blog.

pjsier:
Ohhh that makes more sense. In that case this works for me, other than changes that I'm guessing are on your radar like translation. I'll also include some version of `national.csv` in the standard tool data so that it'll be cached

james-minton:
Love it. @noelelusano here's the tool I was telling you about. Any feedback?

I like having the title in with the tool, especially in case this travels outside the blog ... is that spelled out completely, or done using variables?

pjsier:
Added the national file to the build steps, so `nationalDataUrl` can be changed to `https://evictionlab.org/data/us/national.csv`

noelelusano:
@lane @james-minton looks good to me. Can we apply the same padding that exists between the x baseline and the numbers below it, to the y axis and the numbers to the left of it (looks tight on left. We had same issue with graph on the site)

Lane:
- Shifted y axis labels to the left
  - Added Spanish translations via query param
  - Updated environment URLs

image

Should be ready for merge.

pjsier:
In that screenshot, shouldn't the rate text in the title be translated as well?

Lane:
Yep! Just noticed that too.  One sec, will fix.

james-minton:
Are we good to share with Research? Thanks!

PR #1163, opened 2018-04-28
Fix tertiary event

### LINKED ISSUE #1162: Tertiary Selection GA event not including location label
Label data is blank for Tertiary Selection events even though it's filled in for Primary and Secondary

pjsier:
Closes #1162. Not sure how to test it, but it looks like `tertiaryLocationSelection` should be `tertiaryLocation` based on the custom dimensions set up in Google Analytics and the other events


Lane:
Good catch! Just checked out GTM setup and `tertiaryLocation` is correct.

Issue #1212, opened 2019-11-20
Location-cards adjust info tooltip trigger positioning


amygroshek:
Please shift the "?" upward to be vertically centered relative to "max /min."

<img width="199" alt="Screen_Shot_2019-11-18_at_2 10 01_PM" src="https://user-images.githubusercontent.com/746182/69207256-bd377080-0b14-11ea-8542-dd403ce66ff8.png">

amygroshek:
<img width="611" alt="Screen Shot 2019-11-21 at 12 16 22 PM" src="https://user-images.githubusercontent.com/746182/69364979-fed13400-0c58-11ea-9821-ca6f6358ff0f.png">

@james-minton Pushed up. Will merge later today.

PR #1217, opened 2019-11-21
Adjusts positioning of CI info icon in map cards, #1212

amygroshek:
@james-minton Here's the adjusted positioning. Let me know if you want more adjustments.

Screen Shot 2019-11-20 at 6 14 13 PM


james-minton:
Great, thanks! Can we do a 3px nudge rightward (tested this in Inspector) to add a little padding and then merge?

PR #1235, opened 2020-01-10
Security Updates

### LINKED ISSUE #1231: Address all dependency/security issues across all EL repos

Lane:
Closes #1231 by updating dependencies.

There are still some unresolved vulnerabilities after these updates, but opting to leave them unresolved because:

- they are low or moderate vulnerabilities
- the vulnerabilities are in dev dependencies and are not part of the public facing build
- significant resources will need to by dedicated to fully resolve these vulnerabilities as we will need to perform major version upgrades to the angular CLI and angular core.

An additional change in this PR is including the `angular-d3-graph` module locally in `/src/app/graph`.  This has been done because initially it was thought that we would use this module in other projects, but that is not the case.  Keeping it in this repository will make it easier for future updates to functionality and dependencies.


amygroshek (approved):
All makes sense. Looks good to me. No reason to upgrade Angular at this point.

Good idea moving the graph into the main repo. That will definitely make dev work on that component faster.

Issue #1228, opened 2020-01-06
Condense multiple chart tooltips into one


james-minton:
Change multiple tooltips on chart to a single tooltip containing all locations and their data, and eliminate the secondary readout (on the side of the chart). Cf. Google Data Studio:

<img width="327" alt="Screen Shot 2020-01-06 at 1 36 31 PM" src="https://user-images.githubusercontent.com/33635008/71839622-9ebc1300-3089-11ea-9399-872c958cd9c0.png">

Next steps

- [ ] Design mockup
- [ ] Prototype implementation
- [ ] Launch

Peterlollo:
Closed via PR https://github.com/EvictionLab/eviction-maps/pull/1242

PR #1244, opened 2020-01-30
Implements addl graph padding for CI if CI display is enabled

amygroshek:


Peterlollo (approved):

amygroshek:
Relates to but not a dependency of #1228

